PR #17, opened 2015-12-29
Update go get exmaple

marcosnils:
Change go get example so apex gets automatically installed after downloading. Otherwise if getting `/apex/apex` manual build / installation is necessary.


tj:
ah bummer, tricked myself into thinking that worked

Issue #6, opened 2015-12-20
Delete support


tj:
w/confirmation via https://github.com/segmentio/go-prompt, possibly better to just not allow this haha, but people should restrict access via IAM anyway

mthenw:
If you haven't started that I can take care of this feature.

tj:
That'd be great! It'll probably be a few days until I can hack on this some more

PR #29, opened 2015-12-31
Undeploy function

mthenw:
Fixes https://github.com/apex/apex/issues/6


tj:
Nice LGTM! Only nit is undeploy sounds a little weird to me for some reason. I like 'delete' since it's the same char-length and starts with 'd' haha /OCD

mthenw:
Renamed :)

tj:
:+1:

PR #31, opened 2016-01-04
Function Delete unit tests

mthenw:
Instead of using testify for mocking I used [gomock](https://github.com/golang/mock) for both mocking and generating mock for lambda API. Though I still use testify for assertions because it's more concise.

Let me know if it's ok for you.


tj:
awesome! thanks

PR #36, opened 2016-01-05
Rollback

### LINKED ISSUE #5: Rollback command
```
$ apex rollback
$ apex rollback --version N
```

mthenw:
Closes #5

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -39,6 +39,9 @@ const (
 	DryRun                         = "DryRun"
 )
 
+// Current alias name.
+const CurrentAlias = "current"
+
 // InvokeError records an error from an invocation.
 type InvokeError struct {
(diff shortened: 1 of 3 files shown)


tj:
lgtm!

function/function.go
```diff
@@ -264,6 +283,63 @@ func (f *Function) Invoke(event, context interface{}, kind InvocationType) (repl
 	return reply, logs, nil
 }
 
+// Rollback the function to the previous or specified version.
+func (f *Function) Rollback(version ...string) error {
```
tj: nitpicky but I'd almost prefer RollbackVersion + Rollback even if there's the extra request (assuming it doesn't take long) just to separate the two paths

mthenw: Sure, in a moment.

PR #54, opened 2016-01-07
Function prefixing

### LINKED ISSUE #51: Function prefixing
As @mthenw mentioned they'll just clash if you have multiple projects.

mthenw:
Closes #51

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -62,20 +62,22 @@ type Config struct {
 }
 
 // Function represents a Lambda function, with configuration loaded
-// from the "lambda.json" file on disk. Operations are performed
+// from the "function.json" file on disk. Operations are performed
 // against the function directory as the CWD, so os.Chdir() first.
 type Function struct {
 	Config
-	Path    string
-	Verbose bool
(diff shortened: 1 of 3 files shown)


function/function.go
```diff
@@ -418,3 +421,7 @@ func (f *Function) ZipBytes() ([]byte, error) {
 	f.Log.Infof("created zip (%s)", humanize.Bytes(uint64(len(b))))
 	return b, nil
 }
+
+func (f *Function) lambdaName(projectName string) {
```
ctrlrsf: Minor nitpick, but could this function be renamed to something with more meaning? I couldn't tell what it was doing when it was called up in Open().

mthenw: Yeah, I need to add description also.

tj: might be a nice tiny tweak to just treat it as an arbitrary prefix instead of projectName specifically, not a big deal though. looks like it needs to be rebased with master for whatever reason

mthenw:
@tj rebased

PR #58, opened 2016-01-08
add CloudWatchLogs tailing

tj:
CC @mthenw 

CloudWatchLogs is pretty slow, they take a good ~5s to come in but whatever haha, works.

### DIFF
--- a/cmd/apex/apex.go
+++ b/cmd/apex/apex.go
@@ -11,11 +11,13 @@ import (
 	_ "github.com/apex/apex/runtime/python"
 
 	"github.com/apex/apex/function"
+	"github.com/apex/apex/logs"
 	"github.com/apex/apex/project"
 	"github.com/apex/log"
 	"github.com/apex/log/handlers/text"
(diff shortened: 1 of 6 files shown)


cmd/apex/apex.go
```diff
@@ -252,3 +258,23 @@ func build(project *project.Project, name []string) {
...
 }
+
+// tail outputs logs with optional filter pattern.
+func tail(name []string, filter string) {
+	service := cloudwatchlogs.New(session.New(aws.NewConfig()))
+
+	l := logs.Logs{
+		Service:       service,
+		Log:           log.Log,
+		LogGroupName:  fmt.Sprintf("/aws/lambda/%s", name[0]),
```
mthenw: probably you need to use prefixed name here.
also assuming that I won't pass name here, I will get logs from all project. not the current one, right?

tj: For right now the name is required, not sure why docopt still creates a string slice but it's required ATM

tj:
rebased+prefix in a hacky way for now haha

Issue #52, opened 2016-01-07
Config validation & defaults


tj:
We should probably validate the config on Open(), via gopkg.in/validator.v2 or similar. Even just required fields would be fine for now. 

- [ ] Add required fields
- [ ] Add defaults to "memory" of ??
- [ ] Add defaults to "timeout" of ??
- [ ] Change "role" to be optional once we support it at the project level
- [ ] Update https://github.com/apex/apex/wiki/function.json

mthenw:
I will add required fields with #40

mthenw:
In terms of defaults I would use AWS Lambda defaults which is memory: 128 and timeout: 3

mthenw:
1, 2, 3, 4 done.

tj:
updated wiki!

mthenw:
IMHO https://github.com/apex/apex/wiki/project.json should also be updated :)

PR #60, opened 2016-01-11
Inherit function config from project config

mthenw:
I was expecting much more code but turns out that it's enough to have default values and support configs inheritance.

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -23,6 +23,10 @@ var ErrNotFound = errors.New("project: no function found")
 type Config struct {
 	Name        string `json:"name" validate:"nonzero"`
 	Description string `json:"description"`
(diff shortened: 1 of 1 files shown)


mthenw:
I think it solves https://github.com/apex/apex/issues/52.

project/project.go
```diff
@@ -16,6 +16,13 @@ import (
 	"github.com/tj/go-sync/semaphore"
 )
 
+const (
+	// Memory defines default memory value for every function in a project
+	Memory = 128
```
tj: `DefaultMemory` would be rad but other than that LGTM!

mthenw: Done

tj:
boom!

PR #62, opened 2016-01-12
add initial --dry-run implementation. Closes #47

### LINKED ISSUE #47: Add --dry-run
Maybe. It would probably be cleanest to just swap out the lambda service with a dummy for some operations.

tj:
Need to play with output some more, not happy with it, and finish the inline TODOs. Currently looks like:
screen shot 2016-01-12 at 3 40 06 pm

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -30,6 +30,7 @@ Currently supports:
 - Function rollback support
 - Tail function CloudWatchLogs
 - Concurrency for quick deploys
+- Dry-run to preview changes
 
 ## Example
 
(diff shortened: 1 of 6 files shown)


mthenw:
awesome!

tj:
new look. might lock concurrency later so they're in order, that would be a bit easier to read

screen shot 2016-01-13 at 10 20 03 am

tj:
k ready for review!

mthenw:
LGTM!

PR #64, opened 2016-01-13
Infer name

mthenw:
Actually, we could do the same with project name :) Will do that in separate PR.


tj:
totally, project be a little more sketchy depending on the dir they check out but functions would be no problem

tj:
might almost be worth removing the name field from functions, if they were every different than the dir that would be confusing

Issue #73, opened 2016-01-16
Top-level function name support


pilwon:
It'd be great if we can also deploy and manage top-level named lambda functions with `apex`.

(Example: `{{functionName}}` rather than `{{projectName}}_{{functionName}}`)

Use cases:
* want to ~~deploy~~ colocate all lambda functions (their source codes) in a single `apex` project repo
* want to have more control with the function names used

tj:
Hmmm changing the template could cause collisions. For example later when we support deps properly you might have project A with `{{functionName}}` and project B with the same, then you're screwed. As far as "want to deploy all lambda functions from a single apex project" goes you can deploy them all with `apex deploy`, or did you mean something else?

pilwon:
@tj I meant more about the way one organizes source code for lambda functions.

I personally prefer to use `apex` by setting up a separate repo dedicated for lambda deployments for all my projects (and their functions) deployable to a single AWS account. `apex` shim or pure `lambda` handlers would simply pass `event` and `context` to packages from other repos that have real logic. The benefit of this approach is `apex`-specific repo stays lightweight while the projects with actual logic are decoupled from `apex` or `lambda` as its primary deployment destination.

Namespacing with `{{projectName}}_{{functionName}}` is a perfect default to prevent name collisions and I support keeping this default unchanged. My suggestion is to add an additional option to drop `{{projectName}}_` prefixes.

tj:
Ahh I see, maybe a more ideal thing would be to have nested projects for that. We don't have support yet but there's no reason you couldn't have multiple in one repo. The "master" project would just have the others as local deps. Any thoughts on that? I'm not against the function name thing but that seems like it would be more of a quick fix, if it doesn't have any other use-cases.

pilwon:
@tj Being able to nest projects would be a quite useful project organization feature, however all functions having the same `{{masterProjectName}}_` prefix and no option to opt out from it seems somewhat limiting.

To elaborate on my second use case above, `apex`-deployed functions may not mix well with naming conventions adopted by existing or new functions managed outside `apex` (manually or via alternative tools). Instead of ruling everything with `apex`'s default, having an option to override the default either at project-level or function-level seems like the simplest solution to this problem. An alternative solution is allowing users to define a custom function name template in `project.json`. It's more powerful but also more complicated.

tj:
[...]

PR #75, opened 2016-01-16
Add name template

tj:
As discussed in #73. This adds the new `nameTemplate` project-level property, with default of: "{{.Project.Name}}_{{.Function.Dir}}". The function's "name" property now overrides this project-level value, which I think is nicer semantically, the name field is useless if it matches the directory.


mthenw:
LGTM :+1:

PR #76, opened 2016-01-16
add project nameTemplate support. Closes #73

### LINKED ISSUE #73: Top-level function name support

tj:
As discussed in #73. This adds the new nameTemplate project-level property, with default of: "{{.Project.Name}}_{{.Function.Dir}}". The function's "name" property now overrides this project-level value, which I think is nicer semantically, the name field is useless if it matches the directory.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -24,6 +24,7 @@ Currently supports:
 - Supports languages Lambda does not natively support via shim, such as Go
 - Binary install (useful for continuous deployment in CI etc)
 - Project level function and resource management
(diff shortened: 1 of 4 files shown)


mthenw:
LGTM :+1:

PR #87, opened 2016-01-18
Add actual function name in function log

pilwon:
This PR adds to function log the actual function name rendered by function name template.


tj:
I thinkkkk we should keep functionName to debug output maybe. At least in my case I don't care what they are because I know how apex handles them

pilwon:
@tj Yeah I agree it's better not to show it by default. Updated PR that moves logging to debug output.

function/function.go
```diff
@@ -126,6 +126,9 @@ func (f *Function) Deploy() error {
 // DeployCode generates a zip and creates or updates the function.
 func (f *Function) DeployCode() error {
 	f.Log.Info("deploying")
+	logDeployed := func() {
```
tj: I'd prefer just the one Debugf call right after Info. Sorry I'm super nitpicky :D

pilwon: @tj Not a problem at all. I tried that first but printing "deploying" then "deploying (foo)" right after seemed a bit redundant. What debug message do you think is appropriate to go here?

tj: haha yea agreed, definitely looks weird. personally I'm fine with only seeing the "real" function name via --dry-run but not against it being in debug logs either

tj:
screen shot 2016-01-17 at 6 31 58 pm

tj:
should look a bit better when we move most of those to debug-level haha. thanks for the suggestion!

pilwon:
@tj Your change looks great to me.

PR #88, opened 2016-01-18
Added --versions support to apex list command. Closes #80

### LINKED ISSUE #80: Version listing command
It would be nice to have a version listing command, especially for performing rollbacks. Maybe display the time of deployment etc. Something like:

- `apex list --versions`
- `apex list --versions <name>`

TODO: decide on output

pilwon:
Added version listing command as described in #80

```sh
$ apex list --versions

  - foo (nodejs) - $LATEST [2016-01-18T01:31:50.501+0000]
  - bar (golang) - $LATEST [2016-01-18T01:31:50.501+0000]

```


tj:
woot! I was originally thinking `--versions` would output all versions, but I guess that would be super verbose, even for a single function. I think we can take off the version flag for now, since the default output should at least show the current version

cmd/apex/apex.go
```diff
@@ -149,11 +150,37 @@ func main() {
...
 // list functions.
-func list(project *project.Project) {
-	// TODO(tj): more informative output
+func list(project *project.Project, versions bool) {
+	var infos []map[string]string
+	if versions {
+		var wg sync.WaitGroup
+		wg.Add(len(project.Functions))
+		infos = make([]map[string]string, len(project.Functions))
+		for i, fn := range project.Functions {
```
tj: I know AWS throttles but I'm not positive if the go-sdk retries or not (by default at least). Need to confirm, but we might need to place a limit here

cmd/apex/apex.go
```diff
@@ -149,11 +150,37 @@ func main() {
...
+				}
+			}(i)
+		}
+		wg.Wait()
+	}
+
 	fmt.Println()
-	for _, fn := range project.Functions {
-		fmt.Printf("  - %s (%s)\n", fn.Name, fn.Runtime)
+	for i, fn := range project.Functions {
```
tj: Fetching the info concurrently will probably be a common use-case, it would be sweet to have `Project.FunctionConfig()` that returns a struct similar to the ./log/log.go package, closing the chan if there's an error. That'd help get non-cli related logic out as well which is always a bonus

cmd/apex/apex.go
```diff
@@ -149,11 +150,37 @@ func main() {
...
+		infos = make([]map[string]string, len(project.Functions))
+		for i, fn := range project.Functions {
+			go func(i int) {
+				defer wg.Done()
+				info, err := fn.Info()
+				if err != nil {
+					return
+				}
+				infos[i] = map[string]string{
+					"Version":      *info.Configuration.Version,
```
tj: We're going to be using the `Function.CurrentAlias` to represent the current version. We don't have support for multiple aliases yet but maybe we should just fetch those and do `<alias>=<version>` in the output now that I think about it

pilwon:
@tj Thanks for your feedback. I took a stab at this issue purely out of fun. I guess it's better to wait until we solidify how we exactly want it to look in #80. Please feel free to close this PR, otherwise I will revisit this later when we have something more solid.

tj:
sound good I'll put some more thought into the output and update the ticket!

PR #94, opened 2016-01-18
Check if function exists before deleteing. Closes #65

### LINKED ISSUE #65: Fetch list of functions on delete
Instead of assuming all of the local functions are deployed, we should attempt fetching them first. Maybe a `Function.Exists()` method would be nice here as well.

mthenw:


function/function.go
```diff
@@ -182,7 +182,7 @@ func (f *Function) Delete() error {
 
 // Info returns the function information.
 func (f *Function) Info() (*lambda.GetFunctionOutput, error) {
-	f.Log.Debug("fetching config")
+	f.Log.Debug("fetching info")
```
tj: I'd +1 changing Info to something GetConfig or something less vague haha

project/project.go
```diff
@@ -173,10 +175,19 @@ func (p *Project) Delete(names []string) error {
 		fn, err := p.FunctionByName(name)
 
 		if err == ErrNotFound {
-			p.Log.Warnf("function %q does not exist", name)
-			continue
+			return fmt.Errorf("function %q does not exist in project", name)
+		}
+
+		if _, err := fn.Info(); err != nil {
+			if awserr, ok := err.(awserr.Error); ok && awserr.Code() == "ResourceNotFoundException" {
```
tj: I think in this case we should just ignore since it's not really a big deal, you'd still want to delete the others. Maybe an info or debug level saying "%s has not been deployed"

mthenw: Sure

project/project.go
```diff
@@ -173,10 +175,19 @@ func (p *Project) Delete(names []string) error {
 		fn, err := p.FunctionByName(name)
 
 		if err == ErrNotFound {
-			p.Log.Warnf("function %q does not exist", name)
-			continue
+			return fmt.Errorf("function %q does not exist in project", name)
```
mthenw: @tj what about this one. I'm not sure, would be exit here?

PR #102, opened 2016-01-18
Refactored logs & added -f / --follow support. Closes #57

### LINKED ISSUE #57: Don't tail logs by default
Add -f for tailing, or a flag for the opposite but copying tail -f is probably more intuitive. Default to the most recent N items? Add an `-n` flag as well?

pilwon:
* Renamed `project.name()` to `project.RenderFunctionName()` (such that `logs` package can call this function)
* Refactored `logs.Logs` to take Project and FunctionName (@tj's TODO comment)
* Fixed bug in group name computation not respecting `nameTemplate` (bug fix)
* Added `-f`/`--follow` flag support to `apex logs`. (doesn't follow by default) (issue #57)

### DIFF
--- a/cmd/apex/apex_logs.go
+++ b/cmd/apex/apex_logs.go
@@ -49,15 +49,12 @@ func logsCmdRun(c *cobra.Command, args []string) {
 	lv := &logsCmdLocalValues
 	service := cloudwatchlogs.New(pv.session)
 
-	// TODO(tj): refactor logs.Logs to take Project so this hack
-	// can be removed, it'll also make multi-function tailing easier
-	group := fmt.Sprintf("/aws/lambda/%s_%s", pv.project.Name, lv.name)
-
 	l := logs.Logs{
-		LogGroupName:  group,
-		FilterPattern: lv.Filter,
 		Service:       service,
 		Log:           log.Log,
+		Project:       pv.project,
+		FunctionName:  lv.name,
+		FilterPattern: lv.Filter,
 	}
 
 	for event := range l.Tail() {
--- a/cmd/apex/apex_logs.go
+++ b/cmd/apex/apex_logs.go
@@ -3,10 +3,8 @@ package main
 import (
 	"fmt"
 
-	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
 	"github.com/spf13/cobra"
 
-	"github.com/apex/apex/logs"
 	"github.com/apex/log"
 )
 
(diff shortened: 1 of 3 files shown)


cmd/apex/apex_logs.go
```diff
@@ -49,22 +51,28 @@ func logsCmdRun(c *cobra.Command, args []string) {
...
-		log.Fatalf("error: %s", err)
+	if lv.Follow {
+		for event := range l.Tail() {
+			fmt.Printf("%s", *event.Message)
+		}
+		if err := l.Err(); err != nil {
+			log.Fatalf("error: %s", err)
+		}
+	} else {
+		events, err := l.Fetch()
```
tj: we could probably get away with the same api for both, and one just closes the channel after however many events instead of tailing. not a huge huge deal but hey

logs/logs.go
```diff
@@ -67,3 +102,16 @@ func (l *Logs) loop(ch chan<- *cloudwatchlogs.FilteredLogEvent) {
 func (l *Logs) Err() error {
 	return l.err
 }
+
+// logGroupName returns CloudWatch log group name.
+func (l *Logs) logGroupName() (string, error) {
```
tj: IMO it's a bit cleaner to leave Project/Function completely out of the picture if we don't really need them. In this case I think just passing the LogGroupName in is kinda nice that way

cmd/apex/apex_logs.go
pilwon: @tj I actually first tried reusing existing `loop()` function but it seemed a bit hackish to me. (especially the way `Tail()` returns error to the caller) Your call though, I'll be happy to change it so we reuse more code.

logs/logs.go
pilwon: @tj Ah gotcha. I sort of blindly followed your TODO comment there haha

> // TODO(tj): refactor logs.Logs to take Project so this hack can be removed, it'll also make multi-function tailing easier

tj: hahah shit, my bad. I guess it's not so bad, maybe slightly more elegant would be to have `Project.Logs() *Logs` which computes the name and leaves the rest to Logs. `Project.Logs().Tail()`

cmd/apex/apex_logs.go
tj: yea you're right, nevermind!

logs/logs.go
pilwon: @tj Agreed. I'll add commit for that soon.

pilwon:
@tj Moved log group name logic to `project` pkg so we have `project.Logs(session, name, filter).Tail()` API as per https://github.com/apex/apex/pull/102#discussion_r50056954

PR #103, opened 2016-01-19
Changed "apex --version" to "apex version". Closes #101

pilwon:
* Dropped `-V` and `--version` persistent flag
* Added `apex version`

Closes #101

PR #104, opened 2016-01-19
Skip persistent value prerun for version and wiki

pilwon:
Both `apex version` and `apex wiki` do not rely on persistent flags nor `Project`, so we can skip the prerun on persistent values. This allows users run both commands anywhere, even at a non-apex project root.

EDIT: A better way to implement this is overriding `PersistentPreRun` at the command level. I will refactor after PR #103 is merged to the `master` branch.


tj:
oh nice hahah just mentioned this in the other issue :D

PR #108, opened 2016-01-19
Replaced prerun functions with error versions ("PreRunE")

pilwon:
Replaced `PreRun` with `PreRunE` (the one returns error) such that errors from prerun functions are propagated so it can be printed with the usage.

--

Before:
```
$ apex invoke
error: Missing name argument
```

After:
```
$ apex invoke
Usage:
  apex invoke <name> [flags]

Examples:
  Invoke a function with input json
  $ apex invoke foo < request.json

Flags:
  -a, --async   Async invocation
  -h, --help    help for invoke
  -L, --logs    Print logs

Global Flags:
  -C, --chdir string       Working directory
  -D, --dry-run            Perform a dry-run
  -e, --env value          Environment variable (default [])
  -l, --log-level string   Log severity level (default "info")
  -v, --verbose            Output verbose logs
  -y, --yes                Automatic yes to prompts

error: Missing name argument
```


tj:
nice! do they have a template for this too? might be nicer to have:

```
$ apex invoke

error: Missing name argument

Usage:
  apex invoke <name> [flags]
...
```

pilwon:
@tj Yeah cobra is super flexible with what you want to output. Do you just want to swap the order?

tj:
yep! I think that would help a lot

pilwon:
@tj Big downside of providing a custom template though is this project needs to track changes they make to the [built-in template](https://github.com/spf13/cobra/blob/master/command.go#L265-L289)... Variable names used in template can change more quickly than their API. Do we still want it?

tj:
Hmm I don't see anything there about errors

tj:
Looks like maybe we should add remove the error silencing: 
https://github.com/spf13/cobra/blob/master/command.go#L665-L675

pilwon:
@tj Hmm this one is trickier than I thought. So disabling `SilenceErrors` seems like the only way to print error before usage, but it'll print twice again (top & bottom). The bottom one comes from the `main.go`'s `fmt.Fprintf` which I don't think is a good idea to remove.

tj:
hmm couldn't we remove main.go's one? depends on Cobra's implementation there I guess, but if it's handling help output etc we should be able to either ignore it or assert non-cobra errors or similar

tj:
closing for now since it's getting stale, but I do like this better, we just need to figure out the error part there

PR #113, opened 2016-01-20
Introducing CloudWatch metrics retrieval for functions

### LINKED ISSUE #96: Metrics support
I could offer to implement CloudWatch Metrics for Lambda functions. Sometimes it is quite nice to check:
* Invocations
* Errors
* Duration
* Throttles

e.g.:

```shell
$ master: apex metrics foo --today
Invocations: 	3
Errors: 		1
Duration: 		123 ms
Throttles: 		0
```

```shell
$ master: apex metrics foo --today --invocations --timeline
{ "Datapoints": [{
  "Sum": 1,
  "Timestamp": "2016-01-18 15:00:00 +0000 UTC",
  "Unit": "Count"
},{
  "Sum": 2,
  "Timestamp": "2016-01-18 16:00:00 +0000 UTC",
  "Unit": "Count"
}]
```

johannesboyne:
resolve #96 - Feedback is much appreciated.


cmd/apex/apex_metrics.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+type MetricsCmdLocalValues struct {
+	Start string
+	End   string
+	name  string
+}
+
+const metricsCmdExample = `  Retrieves the CloudWatch metrics for a function
+  $ apex metrics foo
```
mthenw: would be great if you could show two examples, one without --start/endDate and describe it as metrics from last 24 hours, and the second one with --start/endDate as it is now.

johannesboyne: @mthenw do you mean add additional information, like commends at the end:
```shell
$ apex metrics foo #last 24 hours
$ apex metrics foo --start "18/01/2016 10:00" --end "19/01/2016 22:00" #custom timeframe
```
because otherwise is L20 the 24h and L21 the custom timeframe example shown.

mthenw: Maybe:
```
const metricsCmdExample = `  Retrieves the CloudWatch metrics for a function from the last 24 hours
  $ apex metrics foo 

  Retrieves the CloudWatch metrics for a function from custom timeframe
  $ apex metrics foo --start "18/01/2016 10:00" --end "19/01/2016 22:00"`

File: metrics/metrics.go [hunk 2]
```diff
@@ -0,0 +1,113 @@
...
+func aggregate(values []*cloudwatch.Datapoint) int {
+	aggregated_sum := 0.0
+	for _, dp := range values {
+		aggregated_sum += *dp.Sum
+	}
+	return int(aggregated_sum)
+}
+
+// prettify formats the output.
+func (mc *MetricCollector) prettify(ch <-chan Metric) {
```
tj: personally I prefer output to be in cmd code, not a big deal but something to consider!

metrics/request.go
```diff
@@ -0,0 +1,79 @@
...
+	"sync"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/cloudwatch"
+	"github.com/aws/aws-sdk-go/service/cloudwatch/cloudwatchiface"
+)
+
+// request does the final CloudWatch API call
+func (mc *MetricCollector) request(svc cloudwatchiface.CloudWatchAPI, in string, out chan Metric, mutex *sync.Mutex) {
```
tj: one file per struct is usually nicer I find, I was looking for this method up above

johannesboyne: that's true, I extracted it to avoid a long file but let me change it, one shouldn't have to look for struct methods.

metrics/request.go
```diff
@@ -0,0 +1,79 @@
...
+	mc.Collected += 1
+	mutex.Unlock()
+
+	mutex.Lock()
+	curCollected := mc.Collected
+	mutex.Unlock()
+
+	if curCollected == len(mc.Metrics) {
+		out <- *m
+		close(out)
```
tj: not 100% sold on the chan usage with collect/gen/request, the fn creating the channel should typically be the one closing it

metrics/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+			fmt.Fprintf(w, "\033[%dm%s:\033[0m\t%v\n", 37, mm.Name, mm.Count)
+		}
+	}
+	w.Flush()
+}
+
+// collect starts a new cloudwatch session and requests the key metrics.
+func (mc *MetricCollector) collect(in <-chan string) <-chan Metric {
+	out := make(chan Metric)
+	var mutex = &sync.Mutex{}
```
tj: If the request change below is made we can remove the Mutex

tj:
other than the notes LGTM!

johannesboyne:
@tj I've removed the `mutex` but had to introduce a `WaitGroup` I'm not quite sure how one could make it parallel without it. And I've combined the two files ;) 
How about the output, shall I move it to `cmd/apex/apex_metrics.go`?
Furthermore I guess semaphore had a little hick-up while pulling deps.

tj:
ah if you rebase the semaphore thing will go away 590e0e44263bfa06db544bc8183020418325fb26

johannesboyne:
@tj I changed all mentioned and I've rebased everything into a single commit. Anything else to do?

tj:
awesome! mergin

metrics/metrics.go
```diff
@@ -0,0 +1,135 @@
...
+
+			var duration int
+			switch x := mc.EndDate.Sub(mc.StartDate).Hours(); {
+			default:
+				duration = 1 // hourly
+			case x > 24:
+				duration = 24 // daily
+			}
+
+			period := time.Duration(duration) * time.Hour
```
tj: how come we do this stuff instead of just passing in the seconds for the range specified?

johannesboyne: Because of the 1,440 data points limit (http://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/CloudWatch.html#GetMetricStatistics-instance_method).

tj: hmm seems like it's maybe not working as advertised, I've invoked this function a few times ~10 minutes ago with no new metrics

PR #120, opened 2016-01-21
Fix formatting

farazfazli:


mthenw:
There is no need to do that because we don't edit that file manually. This file is generated by go-bindata.

PR #123, opened 2016-01-21
add .apexignore support. Closes #69

### LINKED ISSUE #69: Add support for excluding files from the zip
Just causes bloat if there are extra files. A .gitignore style file would be great, inclusive might be nice as well. Needs more thought as to what this would be and where it would live. If it's in JSON we can use the project.json inheritance etc instead of maintaining it per-function.

mthenw:
It uses shell file name pattern matching to exclude files from zip. 

`.apexignore` in function's dir is concatenated with project's dir. All patterns uses function's dir as a root. 

I will add wiki page tomorrow.


function/function.go
```diff
@@ -101,6 +103,20 @@ func (f *Function) Open() error {
 	}
 	f.runtime = r
 
+	ignorePath := filepath.Join(f.Path, ".apexignore")
+	if _, err := os.Stat(ignorePath); err == nil {
```
tj: not a big deal but might as well have the ReadFile() here & os.IsNotExist()

mthenw: sure, will fix that

tj:
dope. wonder if we should maybe move the bulk of the file loading into utils

mthenw:
Actually I was thinking the same because it's too big. Something like:

`utils.LoadFile(path string, ignoredpatterns []string) (map[string]*os.File, error)`?

mthenw:
@tj, kill it with merge :)

utils/utils.go
```diff
@@ -11,3 +13,44 @@ func Sha256(b []byte) string {
...
+
+// LoadFiles return filtered map of files; for filtering it uses shell file name pattern matching
+func LoadFiles(root string, ignoredPatterns []string) (map[string]*os.File, error) {
+	files := make(map[string]*os.File)
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.Mode().IsRegular() {
```
tj: symlinks would probably be good too, I don't use them much personally but hey

tj: nvm they dont even have an IsSymlink haha ;D

PR #124, opened 2016-01-21
add utils.ReadIgnoreFile(dir)

tj:
use utils.ReadIgnoreFile() in Function.Open()

add tests for utils.ReadIgnoreFile()

add .apexignore to feature list

add test fixtures for previous tests


tj:
cc @mthenw

PR #129, opened 2016-01-23
Add - Python runtime support for pip packages in deployment

pfortin-urbn:


runtime/python/python.go
```diff
@@ -25,3 +26,18 @@ func (r *Runtime) Shimmed() bool {
...
+	cmd := "ls | grep -v -E 'main.py|requirements.txt|test.go' | rm -Rf"
+	_, err := exec.Command("bash","-c",cmd).Output()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *Runtime) Build (dir string) error {
+	cmd := exec.Command("pip", "install", "-r", "requirements.txt", "-t", ".")
```
tj: I'm not too familiar with the ecosystem but isn't this just for installing dependencies? There's not really a build step since it's interpreted

runtime/python/python.go
```diff
@@ -25,3 +26,18 @@ func (r *Runtime) Shimmed() bool {
 func (r *Runtime) DefaultFile() string {
 	return "main.py"
 }
+
+func (r *Runtime) Clean(dir string) error {
+	cmd := "ls | grep -v -E 'main.py|requirements.txt|test.go' | rm -Rf"
```
tj: Likewise Clean is to remove build artifacts, so this would just remove your source/dependency declarations locally which is not what people would want

runtime/python/python.go
pfortin-urbn: It's to install the dependancies where AWS lambda wants to see them - this is not the case for normal operations of python where the dependancies are located in the a defined directory 'site-packages'.  Unless I read the AWS docs wrong they want the dependancies installed in the same directory as the python main code.  -- How are you handling node.js dependancies?

runtime/python/python.go
pfortin-urbn: Same comment as above this clears out the stuff that is only there for the zip file.

tj: oops I read the grep wrong, I'll check out the aws docs there I thought requirements.txt was maybe a normal python thing

runtime/python/python.go
tj: For node it's just regular old node_modules so I figured it was similar here

tj: hmm I still think this is probably best left to the developer, it would be very slow to run `npm install` each time, so I'm thinking we should probably treat this the same

tj: I guess to rephrase, the only reason we have this for Go is because it can't work in Lambda without compilation, but I'd like to move that to hooks as well (but that hook will be built in).

runtime/python/python.go
pfortin-urbn: requirements.txt is a normal thing.  test.go was a small test utility to check the code that does the clean and can be removed from that line.  The other python asset needed are what my Build() finc was doing.

pfortin-urbn:
I guess I was under the impression that apex was a more generic tool and not only to be used to deploy go aws lambdas since there was support for node anf python.  But if I have to install manually the dependancies in the proper location then I can use many other tools to deploy to AWS.   The greatest apex selling point for me was the ability to not worry about dependancies, just update them in a file and apex could take care of the deployment correctly for me.  I can delete this PR is you want...

tj:
In general we're just make deploys and managing api gateway (soon) quick and easy. In the case of runtimes that AWS doesn't support we have to get a bit more opinionated, but since they support node/python I don't think we should get too fancy there unless it's something that applies to every single person using the runtime. 

The next release will be tomorrow which adds hooks, so then you'll be able to do this on deploy with a single hook in project.json. I can't speak for the python tooling but like I mentioned with node/npm this would be a really slow thing to add as a default, since you typically only `npm install` if you've made changes to your deps.

cheers

pfortin-urbn:
I have not played with the node deps but I can tell you that where you are suppsed to install the python deps the way AWS wants them you do not normally code or install that way so it is very much part of the deployment and not any part of dev so the install might need to happen each deploy.  Yes I agree it would slow it down but it would also reduce the deployment complexity by a lot - I will see your new version and maybe go from there.

Cheers!

PR #131, opened 2016-01-23
add hook support. Closes #68

### LINKED ISSUE #68: Hook support
Might be nice to have some hooks for builds, linting, running tests, whatever. We could potentially move some things like the shimmed language build commands to this instead, as a default, and remove part of the `runtime.Runtime` interface.

We could then remove `runtime.Runtime` all together if we wanted, and just provide defaults for these properties. Shimmed() would be an option, Handler() an option, DefaultFile() an option, Build() would be a hook, Clean() would be a hook.

This would allow people to extend Apex without contributing Go code in core as well. For example you might even want to shim node code to act over stdio instead, just because it's nicer to work with.

We could also move the Go runtime portion to a different package then, just so they're not conflated, just apex/runtime. 

- [ ] remove runtime.Runtime
- [ ] define defaults
- [ ] refactor how inference works (define the default file detection stuff elsewhere)
- [ ] add "shim" (bool)
- [ ] add "handler" (string)
- [ ] add "hooks" (map of strings)
- [ ] add hooks for pre/post-deploy and others

Pretty sold on this I think. We shouldn't really focus on runtimes, they'll add Go support eventually and same with the others. These could also be inherited from the project level, which would be useful in most cases, for linting, pre-compiling with Babel etc.

tj:
Only adds two hooks for now:

- build (pre-zip)
- clean (post-zip)

We can add more on request, pre/post-deploy might be good for linting/testing but I'm using build and clean here so that the `apex build` command works as well

- [ ] update wiki

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -21,6 +22,7 @@ import (
 	"github.com/jpillora/archive"
 	"gopkg.in/validator.v2"
 
+	"github.com/apex/apex/hooks"
 	"github.com/apex/apex/runtime"
 	"github.com/apex/apex/shim"
 	"github.com/apex/apex/utils"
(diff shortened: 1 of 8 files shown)


function/function.go
```diff
@@ -449,20 +466,64 @@ func (f *Function) Zip() (io.Reader, error) {
...
+// Error string.
+func (e *HookError) Error() string {
+	return fmt.Sprintf("hook %q: %s", e.Name, e.Output)
+}
+
+// RunHook executes hook `name` in the function's directory.
+func (f *Function) RunHook(name string) error {
+	var command string
+
+	switch name {
```
tj: this is kinda lame but executing them unnamed in the log output could be confusing

mthenw: fine for now imho

mthenw:
This is more than great. Thanks!

tj:
woot thanks for the review!

PR #134, opened 2016-01-24
Add awesome animated gif to Readme

farazfazli:


Readme.md
```diff
@@ -3,6 +3,10 @@
 
 Apex is a small tool for deploying and managing [AWS Lambda](https://aws.amazon.com/lambda/) functions. With shims for languages not yet supported by Lambda, you can use Golang out of the box.
 
+<p align="center">
+  <img src="https://github.com/farazfazli/apex/blob/master/assets/terminal.gif" alt="Sublime's custom image"/>
```
tj: src="assets/terminal.gif" :D, it's hard-coded to yours right now

tj: looks low res on retina, I can try one on my machine and see if it's less blurry

farazfazli: Hahaha, yes it is...we can change it after the merge! I will try recording it again.

tj:
out

tj:
hmm that looks blurry too wtf tj/stack's one doesn't look bad

tj:
nvm retina ones are like 30mb

Readme.md
```diff
@@ -3,6 +3,10 @@
 
 Apex is a small tool for deploying and managing [AWS Lambda](https://aws.amazon.com/lambda/) functions. With shims for languages not yet supported by Lambda, you can use Golang out of the box.
 
+<p align="center">
+  <img src="https://github.com/farazfazli/apex/blob/master/assets/terminal.gif" alt="Terminal Example"/>
```
turboMaCk: relative path will be much better I think.

mthenw:
It's quite blurry on Retina, besides do we really need this?

tj:
doesn't matter to me, I kinda like that the node one shows how fast deploys can be with lambda, but otherwise I think it probably works the way people would expect from the docs

farazfazli:
I don't think you can center the image if you use a relative path with markdown.

turboMaCk:
Why not? This does not makes sense. DOM itself is responsible for rendering. It does not not care about src location, just data.

tj:
let's revisit if there's a way to make a non-blurry version that isn't multi-megabytes

farazfazli:
@turboMaCk http://stackoverflow.com/a/12118349

turboMaCk:
```html
<p align="center">
   <img src="assets/terminal.gif" alt="Terminal Example">
</p>
```

would work perfectly.

farazfazli:
Oh okay, @turboMaCk I didn't know you could do that! Thanks! :smiley:

PR #135, opened 2016-01-24
add plugins, replacing runtimes. Closes #130

### LINKED ISSUE #130: Replace runtimes with hooks
This will basically remove the concept of a runtime, moving it into a Plugin which just happens to manipulate Function fields etc.

- [x] remove runtime.Runtime
- [x] add `plugins: ["shim", "coffeescript"]` field
- [x] add default plugins (shim etc)
- [x] add Plugin and plugins.Register
- [x] define defaults
- [x] refactor how inference works (define the default file detection stuff elsewhere)
- [x] add "shim" (bool)
- [x] add "handler" (string)
- [x] move hooks to plugin

This could also be used to inject the shim etc, supporting "shim" as a plugin, instead of a built-in Function concept. It'll just happen to be a plugin that is always enabled. We could maybe move inference here as well. Maybe move hooks to a plugin as well, plenty of things that could be moved out of Function and injected in a more modular way.

tj:
Pretty big PR but this does the following:

- removes runtimes
- adds hooks throughout Function to move some of our logic out into logical "plugins"
- moves most of the old ad-hoc Function logic into these plugins
- adds `"environment": {}` support to project.json and function.json

I'm not completely sold on using constants and a single method for the plugin system but it works for now, might be nicer to use interfaces there but we'd have to move the plugin iteration into separate loops.

They're also not 100% decoupled from Function since we need the Config fields but it's close enough :D. Might add some more tests later haha.. easier to test now at least.


function/function_test.go
```diff
@@ -1,11 +1,18 @@
-package function
+package function_test
```
mthenw: <3

mthenw:
> I'm not completely sold on using constants and a single method for the plugin system

same here but it's fine for now :)

tj:
I'll probably keep the constants and fn.hook() but switch in fn.hook() to call Open/Build/Clean etc, something like that

Issue #139, opened 2016-01-24
fix race in test


tj:
keeps randomly breaking in TestFunction_Open_requireConfigValues

tj:
think it's fine now

mthenw:
nope, still there :)

mthenw:
Not sure how but it looks like fixed. I run test few dozen times without failure. When I tried before about 40% runs failed.

PR #146, opened 2016-01-25
Add support for specifying a duration for the logs cmd

### LINKED ISSUE #136: Add support for specifying the duration used for logs command
`apex logs foo` returns the last minute currently, it would be nice to have `apex logs foo 5m`, `apex logs foo 5h` with time.ParseDuration

johannesboyne:
closes #136 

Feedback is much appreciated.

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -16,13 +16,16 @@ type Logs struct {
 	Log           log.Interface
 	GroupName     string
 	FilterPattern string
+	StartTime     time.Time
+	EndTime       time.Time
 
 	err error
 }
 
 // Fetch log events.
 func (l *Logs) Fetch() ([]*cloudwatchlogs.FilteredLogEvent, error) {
-	start := time.Now().Add(-time.Minute).UnixNano() / int64(time.Millisecond)
+	start := l.StartTime.UTC().UnixNano() / int64(time.Millisecond)
+	end := l.EndTime.UTC().UnixNano() / int64(time.Millisecond)
 
 	l.Log.Debugf("fetching %q with filter %q", l.GroupName, l.FilterPattern)
 
@@ -30,6 +33,7 @@ func (l *Logs) Fetch() ([]*cloudwatchlogs.FilteredLogEvent, error) {
 		LogGroupName:  &l.GroupName,
 		FilterPattern: &l.FilterPattern,
 		StartTime:     &start,
+		EndTime:       &end,
 	})
 
 	if err != nil {
@@ -51,7 +55,8 @@ func (l *Logs) loop(ch chan<- *cloudwatchlogs.FilteredLogEvent) {
 	defer close(ch)
 
 	var nextToken *string
-	start := time.Now().Add(-time.Minute).UnixNano() / int64(time.Millisecond)
+	start := l.StartTime.UTC().UnixNano() / int64(time.Millisecond)
+	end := l.EndTime.UTC().UnixNano() / int64(time.Millisecond)
 
 	l.Log.Debugf("tailing %q with filter %q", l.GroupName, l.FilterPattern)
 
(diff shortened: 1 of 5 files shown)


logs/logs.go
```diff
@@ -16,21 +16,26 @@ type Logs struct {
...
 
-	res, err := l.Service.FilterLogEvents(&cloudwatchlogs.FilterLogEventsInput{
+	fi := &cloudwatchlogs.FilterLogEventsInput{
 		LogGroupName:  &l.GroupName,
 		FilterPattern: &l.FilterPattern,
 		StartTime:     &start,
-	})
+		EndTime:       &end,
+	}
+	res, err := l.Service.FilterLogEvents(fi)
```
tj: nitpick: I prefer just passing the options directly since that's what we've been doing everywhere

project/project.go
```diff
@@ -254,7 +258,7 @@ func (p *Project) FunctionNames() (list []string) {
 }
 
 // Logs returns logs.
-func (p *Project) Logs(s *session.Session, name string, filter string) (*logs.Logs, error) {
+func (p *Project) Logs(s *session.Session, name string, filter string, duration string) (*logs.Logs, error) {
```
tj: this method is getting weird hahaha, we can fix it later though

project/project.go
```diff
@@ -272,6 +276,28 @@ func (p *Project) Logs(s *session.Session, name string, filter string) (*logs.Lo
 		FilterPattern: filter,
 	}
 
+	var parsedStartTime time.Time
```
tj: start/end is nicer than verbose name IMO

project/project.go
```diff
@@ -272,6 +276,28 @@ func (p *Project) Logs(s *session.Session, name string, filter string) (*logs.Lo
...
+	var parsedEndTime time.Time
+
+	if duration != "" {
```
tj: `if duration == ""` with an else negative assertions are weird

project/project.go
```diff
@@ -272,6 +276,28 @@ func (p *Project) Logs(s *session.Session, name string, filter string) (*logs.Lo
...
+		parsedDuration, err := time.ParseDuration(duration)
+		if err != nil {
+			return nil, err
+		}
+		parsedEndTime = time.Now()
+		parsedStartTime = parsedEndTime.Add(-parsedDuration)
+	} else {
+		parsedStartTime = time.Now()
+		parsedEndTime = parsedStartTime.Add(-time.Duration(1) * time.Minute)
```
tj: Are these two backwards? shouldn't the start time be -1M?

johannesboyne:
```
   • START 18:50:58.545935649 +0100 CET
   • END   18:49:58.545935649 +0100 CET
```

tj: but the range is `<start>...<end>` isn't it? not `<end>..<start>`

johannesboyne: > EndTime *int64
> A point in time expressed as the number of milliseconds since Jan 1, 1970 00:00:00 UTC. If provided, events with a timestamp later than this time are not returned.

> StartTime *int64
> A point in time expressed as the number of milliseconds since Jan 1, 1970 00:00:00 UTC. If provided, events with a timestamp prior to this time are not returned.

So, my understand is:
`(past) [----[end]======[start]--------] (now)`

But as I'm no native speaker, maybe it's just confusing :D

johannesboyne: Ok, I thing you're right, "later" is unfortunately in the future...

johannesboyne:
Not quite sure why semaphore failed, the test runs correct on my system, will look into it...

tj:
that's probably just #139 :D

tj:
swapped the ordering there, should be all good now unless the AWS docs are super messed up

PR #149, opened 2016-01-26
update download reference to 0.5.1

sperand-io:
Hey @tj!

I can't seem to find a `latest` url path for release binary downloads. I can imagine keeping this updated in the readme will be pretty annoying / error-prone, so if you have any other ideas for automating bumping it or pointing at a mirror or something lmk and I'm happy to have a go at that too!

Looking forward to making some "real" contributions soon (and using it for some new stuff @ segment!) — awesome progress and momentum so far :clap:!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -8,7 +8,7 @@ Apex is a small tool for deploying and managing [AWS Lambda](https://aws.amazon.
 Download [binaries](https://github.com/apex/apex/releases):
 
 ```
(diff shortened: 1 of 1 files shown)


tj:
that should do for now haha, kinda gross but hey. we also need os/arch detection though if it's going to be more than an example, I'll maybe write a little script tomorrow that is more portable since plenty of binary release projects could use it. Maybe leaving it out and just letting people curl on their own would be fine too, not sure where everyone's curl-foo is it

PR #151, opened 2016-01-26
refactor functions loading. Closes #132

### LINKED ISSUE #132: Don't load all functions for single-function operations
Just realized that we are haha, not a big deal until it starts annoying someone perf-wise

mthenw:
Now, only specified function are loaded (or all if no explicit).

### DIFF
--- a/cmd/apex/apex_delete.go
+++ b/cmd/apex/apex_delete.go
@@ -11,8 +11,6 @@ import (
 
 type DeleteCmdLocalValues struct {
 	Force bool
-
-	names []string
 }
 
 const deleteCmdExample = `  Delete all functions
@@ -25,7 +23,6 @@ var deleteCmd = &cobra.Command{
 	Use:     "delete [<name>...]",
 	Short:   "Delete functions",
 	Example: deleteCmdExample,
-	PreRun:  deleteCmdPreRun,
 	Run:     deleteCmdRun,
 }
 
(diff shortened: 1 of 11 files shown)


cmd/apex/apex_delete.go
```diff
@@ -38,23 +35,18 @@ func init() {
 	f.BoolVarP(&lv.Force, "force", "f", false, "Force deletion")
 }
 
-func deleteCmdPreRun(c *cobra.Command, args []string) {
```
mthenw: @tj @pilwon not sure about that. Can I just remove this method as it doesn't make sense anymore?

cmd/apex/apex_delete.go
```diff
@@ -38,23 +35,18 @@ func init() {
...
+func deleteCmdRun(c *cobra.Command, args []string) {
 	lv := &deleteCmdLocalValues
 
-	if len(args) == 0 {
-		lv.names = pv.project.FunctionNames()
+	if err := pv.project.LoadFunctions(args...); err != nil {
```
mthenw: I've change a logic here a bit. If there is no function with specified name error will be returned. Previously there was a question first. I think it's better because it doesn't give you false impression that you have this function.

tj: :+1: sounds good to me

tj:
awesome <3

Issue #125, opened 2016-01-22
logs missing


cscetbon:
The way the logs tailing currently works appears to miss some of them.

Exemple of the output from `apex logs` :
```
2016-01-22T14:41:38.029Z	50kjimgyfp8h3wjy	start bar
START RequestId: 3e05639e-c116-11e5-b961-0133577b6db1 Version: $LATEST
END RequestId: 3e05639e-c116-11e5-b961-0133577b6db1
REPORT RequestId: 3e05639e-c116-11e5-b961-0133577b6db1	Duration: 10.25 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 9 MB
2016-01-22T14:41:38.962Z	aitexn58yovc65ry	start bar
START RequestId: 3e9a77f7-c116-11e5-a3c3-81a93366183c Version: $LATEST
END RequestId: 3e9a77f7-c116-11e5-a3c3-81a93366183c
REPORT RequestId: 3e9a77f7-c116-11e5-a3c3-81a93366183c	Duration: 1.70 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 9 MB
START RequestId: 3f10449c-c116-11e5-b9d4-f5de8fd301f9 Version: $LATEST
END RequestId: 3f10449c-c116-11e5-b9d4-f5de8fd301f9
REPORT RequestId: 3f10449c-c116-11e5-b9d4-f5de8fd301f9	Duration: 4.12 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 9 MB
START RequestId: 755eb3b8-c116-11e5-a200-dff599438107 Version: $LATEST
END RequestId: 755eb3b8-c116-11e5-a200-dff599438107
REPORT RequestId: 755eb3b8-c116-11e5-a200-dff599438107	Duration: 3.70 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 9 MB
START RequestId: 7c1992de-c116-11e5-946d-756283e77d9d Version: $LATEST
END RequestId: 7c1992de-c116-11e5-946d-756283e77d9d
REPORT RequestId: 7c1992de-c116-11e5-946d-756283e77d9d	Duration: 0.32 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 9 MB
START RequestId: 7c8facc0-c116-11e5-abc3-059416ccd536 Version: $LATEST
[...]

PR #154, opened 2016-01-26
fix log tailing

tj:
fixes log tailing, not sure how EndTime got in there, but we were never updating it, and it's optional so this fixes it


mthenw:
Does it fix https://github.com/apex/apex/issues/125 ?

PR #171, opened 2016-01-28
Vendor all external dependencies

### LINKED ISSUE #169: Vendor deps
Pretty low priority IMO, in practice this isn't a huge problem for something with infrequent releases like Apex. Go 1.6 is coming soon so we'll have the vendor lookup.

One thing I'm not sure of is if you can disable the GOPATH lookup or not, the docs didn't make it clear if ./vendor lookup was basically prepending GOPATH or what, if so I think that's a mistake on their part, having it lookup ./vendor _and_ GOPATH is not good. So if we can have it explicitly use ./vendor only that would be great.

farazfazli:
Closes #169


tj:
I added GO15VENDOREXPERIMENT=1 to Semaphore, still seems unhappy:

```
vendor/github.com/andybalholm/cascadia/parser.go:796: unreachable code
exit status 1
vendor/golang.org/x/crypto/ssh/terminal/terminal.go:744: unreachable code
exit status 1
```

tj:
hopefully they ignore ./vendor in all the tools, we dont need to re-test those

tj:
damn https://github.com/golang/go/issues/11659

farazfazli:
Oh shoot

tj:
I'm meh for now then, not worth all the hacks to strip vendor out

PR #175, opened 2016-01-28
refactor plugins

tj:
This is still annoying but at least now we have real interfaces, I need this
so we can pass the zip to build hooks, letting us skip I/O and do the nodejs
env injection a lot easier.


mthenw:
Love it. Much clearer now.

PR #176, opened 2016-01-28
Add nodejs env injection

tj:
Requires #175 first.


plugins/nodejs/nodejs.go
```diff
@@ -24,3 +43,34 @@ func (p *Plugin) Open(fn *function.Function) error {
...
+		HandleMethod: method,
+	})
+
+	if err != nil {
+		return err
+	}
+
+	fn.Handler = "_index.handle"
+
+	return zip.AddBytes("_index.js", buf.Bytes())
```
mthenw: fingers crossed that someone won't have that file in repo :) It's still not great but maybe we could think about more unique name, _apex_index.js ?

tj: SGTM! might as well

tj:
@mthenw rebased into the two commits

Issue #90, opened 2016-01-18
Pull region from ~/.aws/config


tj:
https://github.com/aws/aws-sdk-go/issues/384

plukevdh:
Is the reference to the issue above to say 'wait for Amazon to decide' or just for reference to the known limitations of the ~/.aws/config`? The `~/.aws/config` _is_ technically for CLI config only. I've typically shelled out to the CLI to ask for this value and, if none is set, use some default (like 'us-east-1'). This behavior (to not include the CLI-configured `~/.aws/config` values) is pretty standard across all of the non-CLI SDKs.

tj:
Either or, I don't mind if we read from it directly or wait for them. It would be great to just swap profiles with a single export

kujohn:
I was planning to implement reading the file directly. But the `~/.aws/config` file format seems to be non-TOML based. Does anyone know what format this is:

```
[default]
region = us-west-2
[profile test]
region = us-east-1
```

tj:
Pretty much any ini parser should work I'd think, TOML is pretty much just that

tj:
46545c7

kujohn:
@tj thanks! Wasn't sure if we had a particular ini parser in mind, we can switch out if required.

PR #179, opened 2016-01-29
support for aws profile switching via aws shared credentials

kujohn:
Support for profile switching discussion from #90


tj:
nice I never noticed they had that :D thanks!

PR #181, opened 2016-01-29
add tfvars output to list command. Closes #155

### LINKED ISSUE #155: Output tfvars file for Terraform
Need to think on integration some more, but this would be pretty minimal. We could have a command or project.json path to a tfvars file that gets populated by Apex so you can reference the function ARNs.

mthenw:
Example output

```
$ apex list --tfvars
apex_function_test="arn:aws:lambda:..."
```

### DIFF
--- a/cmd/apex/apex_list.go
+++ b/cmd/apex/apex_list.go
@@ -2,13 +2,21 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


tj:
:+1: dope, needs rebase I think

PR #184, opened 2016-01-29
add multi-function log support. Closes #159

### LINKED ISSUE #159: Multi-function logs
I'd like to refactor a few things and make it `apex logs [<name> ...]` so all our commands have the same multi-function support.

- [x] refactor log package
- [x] support multiple functions (all by default)
- [x] add flag for duration (default to something larger than what we have now; 30? 10? 5?)
- [ ] update docs

tj:
Pretty big cleanup of how log fetching works, and adds support for multiple to match our other command signatures. Some feedback on a flag name would be great as well, need to finalize that. Duration doesn't sound right when it's in the past. 

Logs are prefixed with the group name, not the function name, so that we can utilize this later for other service logs as well.


cmd/apex/apex_logs.go
```diff
@@ -2,31 +2,37 @@ package main
...
-	name string
+	Duration time.Duration
 }
 
-const logsCmdExample = `  Print logs for a function
-  $ apex logs <name>
+const logsCmdExample = `  Print logs for all functions
+  $ apex logs
+
+	Follow the output
```
mthenw: tab -> space

tj: haha dammit, I always do this

mthenw:
awesome, LGTM :+1:

PR #185, opened 2016-01-29
fix not implementing interface by some plugins

mthenw:


tj:
shit haha, silent errors ftw

tj:
forgot that we can now inject the shim without IO there too

PR #186, opened 2016-01-29
fix hooks error reporting. Closes #160

### LINKED ISSUE #160: Output function name in fn-related errors
Which one was it? Nobody knows!

```
   • deploying                 function=A
   • deploying                 function=B
   ⨯ error: hooks: build: boom
```

It would be nice to have function=X, but since we delegate for errors that wouldn't really work, but I'm cool with `error: <name>: hooks: build: boom`.

Alternatively we always output error logs in-place but that feels gross, or augment errors as FunctionError, which is kind of annoying.

mthenw:


tj:
ah #160 wasn't just for hooks, we have quite a few cases, I think in Project we should wrap/prefix

PR #187, opened 2016-01-29
add function name to error output. Closes #160

### LINKED ISSUE #160: Output function name in fn-related errors

mthenw:
Should be better now :)


mthenw:
Example:
```
⨯ error: function test: build hook: sh: lol: command not found
```

project/project.go
```diff
@@ -164,7 +165,11 @@ func (p *Project) Deploy() error {
 
 			go func() {
 				defer sem.Release()
-				errs <- fn.Deploy()
+
+				err := fn.Deploy()
+				if err != nil {
+					errs <- fmt.Errorf("function %s: %s", fn.Name, err)
```
tj: here we have to handle the nil case as well since we're looping below to "drain" the chan

PR #192, opened 2016-01-29
Add scheduled event support closes #147

### LINKED ISSUE #147: CRON scheduled functions
Seems like they don't provide an API for this yet which is a little weird, hopefully that's coming:
- http://docs.aws.amazon.com/lambda/latest/dg/with-scheduled-events.html

johannesboyne:
I'll continue later and merge the master and than it closes #147 ;-)


johannesboyne:
@tj or @mthenw I've finished the implementation **but** what are your thoughts on deleting scheduled events? For now it is not implemented, just because I didn't want to query the connected CloudWatchEvents each time since most of the time there are no...

tj:
hmm good point, I guess ideally it's more of a terraform thing since it's super CRUD-like, I'll review in a bit!

cmd/apex/main.go
```diff
@@ -7,19 +7,20 @@ import (
...
 	_ "github.com/apex/apex/cmd/apex/invoke"
-	_ "github.com/apex/apex/cmd/apex/delete"
-	_ "github.com/apex/apex/cmd/apex/rollback"
-	_ "github.com/apex/apex/cmd/apex/build"
 	_ "github.com/apex/apex/cmd/apex/list"
-	_ "github.com/apex/apex/cmd/apex/metrics"
 	_ "github.com/apex/apex/cmd/apex/logs"
-	_ "github.com/apex/apex/cmd/apex/docs"
+	_ "github.com/apex/apex/cmd/apex/metrics"
+	_ "github.com/apex/apex/cmd/apex/rollback"
```
tj: wonder if we can impose this order in cobra some other way hahaha `_` + gofmt is annoying there

function/function.go
```diff
@@ -182,11 +188,11 @@ func (f *Function) DeployCode() error {
 	return f.Update(zip)
 }
 
-// DeployConfig deploys changes to configuration.
+// DeployConfig deploys changes to configuration and trigger postDeploy hook
```
tj: was this meant to be here still since there's no hook call here?

function/function.go
```diff
@@ -281,6 +295,8 @@ func (f *Function) Create(zip []byte) error {
 		},
 	})
 
+	f.FunctionArn = *created.FunctionArn
```
tj: I think it might be a little more elegant to just do a GetConfig or similar in the plugin instead of relying on it being set several times throughout here

plugins/cron/cron.go
```diff
@@ -0,0 +1,38 @@
...
+		return p.addCron(fn)
+	}
+	return nil
+}
+
+// addCron builds the Cron Event.
+func (p *Plugin) addCron(fn *function.Function) error {
+	event := &cron.Cron{
+		Name:              "Cron_" + fn.FunctionName,
+		Description:       "Cron_" + fn.FunctionName,
```
tj: maybe "cron_%s" for name and "cron for lambda function %s" for description

hooks/hooks.go
```diff
@@ -10,4 +10,7 @@ type Hooks struct {
 
 	// Deploy command is run after builds and before deploys.
 	Deploy string `json:"deploy"`
+
+	// PostDeploy command is run after deploys.
+	PostDeploy string `json:"postDeploy"`
```
tj: "deployed" for the json part

plugins/hooks/hooks.go
```diff
@@ -46,6 +46,11 @@ func (p *Plugin) Deploy(fn *function.Function) error {
 	return p.run("deploy", fn.Hooks.Deploy, fn)
 }
 
+// PostDeploy runs the "postDeploy" hook commands.
+func (p *Plugin) PostDeploy(fn *function.Function) error {
+	return p.run("postDeployed", fn.Hooks.PostDeploy, fn)
```
tj: "deployed" for the hook name

sources/cron/cron.go
```diff
@@ -0,0 +1,94 @@
...
+				Arn: aws.String(c.FunctionArn),
+				Id:  aws.String("Cron_" + c.FunctionName),
+			},
+		},
+	})
+	if err != nil {
+		return err
+	}
+	if *res.FailedEntryCount > 0.0 {
+		return &failedEntryCountError{"failed entry count > 0"}
```
tj: I'd ditch the custom error since it doesn't have any useful properties and it's not asserted anywhere

sources/cron/cron.go
```diff
@@ -0,0 +1,94 @@
...
+	Expression        string
+	FunctionName      string
+	FunctionArn       string
+	CloudWatchService cloudwatcheventsiface.CloudWatchEventsAPI
+	LambdaService     lambdaiface.LambdaAPI
+}
+
+func (c *Cron) put() (string, error) {
+	res, err := c.CloudWatchService.PutRule(&cloudwatchevents.PutRuleInput{
+		Description:        aws.String(c.Description),
```
tj: all of these except for ENABLED below could be `&c.Description` etc

plugins/cron/cron.go
```diff
@@ -0,0 +1,38 @@
...
+func init() {
+	function.RegisterPlugin("cron", &Plugin{})
+}
+
+// Plugin implementation.
+type Plugin struct{}
+
+// PostDeploy adds the Cron Scheduled Event.
+func (p *Plugin) PostDeploy(fn *function.Function) error {
+	if &fn.Sources != nil && &fn.Sources.Schedule != nil {
```
tj: this part looks like it shouldn't work haha. `Schedule` is a string, and since it's not a pointer this condition could just be `if fn.Sources.Schedule == "" { return nil } return p.addCron(fn)`

sources/cron/cron.go
```diff
@@ -0,0 +1,94 @@
...
+	})
+	if err != nil {
+		return "", err
+	}
+	return *res.RuleArn, err
+}
+
+func (c *Cron) connect(ruleName string) error {
+	res, err := c.CloudWatchService.PutTargets(&cloudwatchevents.PutTargetsInput{
+		Rule: aws.String(ruleName),
```
tj: I'll have to read up on this api a bit, seems crazy that we need three api calls just to add a little cron thing

tj:
sucks TF doesn't have cloudwatchevents, sounds like they're on it https://github.com/hashicorp/terraform/issues/4827, seems like a super terraform-y thing though after seeing the API calls. I'd like to stay away from cross into that zone too much, I just use TF for managing my sources right now

johannesboyne:
Mhm... as a TF "fan" myself I'm with you... just don't want to wait for it, neither do I want to implement all of the CloudWatchEvents stuff for TF.

tj:
I don't know if it's smart for us to throw too much CRUD stuff in Apex though, it'll be awkwardly split between the two. In an ideal world we'd have it all but that's basically just Terraform. The main reason I started Apex was that Lambda function management was really bad in TF, but I don't think that will change haha it's not suited for repeated deploys or workflow related stuff

radeksimko:
Hi,
FYI I just finished the implementation of CloudWatch Events (both rules & targets) in https://github.com/hashicorp/terraform/pull/4986

I'd be more than happy if you tested these resources before they reach upstream. :wink: 

This PR may also be interesting for apex: https://github.com/hashicorp/terraform/pull/4826

tj:
@radeksimko nice! thanks

PR #203, opened 2016-02-01
refactor metrics; add multiple functions metrics. Closes #182

### LINKED ISSUE #182: Multi-function metrics command
Nice to have consistency across all these commands. Change output to look similar to `apex list`.

mthenw:


tj:
niceee, does this output metrics for each? looks like it aggregates them all into one. I think outputting the metrics per-function would be most useful (similar to the `list` output), since you'd usually want to see if any functions are problematic and hitting throttling etc

mthenw:
make sense, fix in a moment

mthenw:
@tj done

tj:
:DDD

PR #205, opened 2016-02-01
add support for AWS_PROFILE region from ~/.aws/config

tj:
this function is getting a little ugly but this adds support for ~/.aws/config's region when using AWS_PROFILE, not just --profile


mthenw:
Could you also update docs?

tj:
think we're good:

```
However, if you do not have a ~/.aws/config file, or "region" is not defined, you should set it with the `AWS_REGION` environment variable.
```

I thought we supported it already

PR #207, opened 2016-02-02
add env populating for python runtime. Closes #202

### LINKED ISSUE #202: Loading environment variables for Python runtime
I use `apex deploy -e log_level=DEBUG App` to deploy an environment variable called `log_level` with my lambda. However it's not loaded in the environment map when I invoke it using `apex invoke App < event.js`. Is it a current limitation of the Python runtime with apex ?

mthenw:
I'm not python expert but it works.

consts FTW! :)

### DIFF
--- a/plugins/env/env.go
+++ b/plugins/env/env.go
@@ -15,6 +15,9 @@ func init() {
 	function.RegisterPlugin("env", &Plugin{})
 }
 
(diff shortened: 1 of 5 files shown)


tj:
sweet LGTM! I'm no python pro either

PR #211, opened 2016-02-02
refactor shim plugin to access zip directly. Closes #190

### LINKED ISSUE #190: Refactor shim plugin to access zip directly
instead of doing IO, just do AddBytes or whatever that method was

mthenw:


tj:
yay so much nicer haha

PR #215, opened 2016-02-04
fix updating configuration. Closes #206

### LINKED ISSUE #206: Updating configuration doesn't work
because we don't use $LATEST. 

AWS Docs:
> If you are using the versioning feature, note this API will always update the $LATEST version of your Lambda function.

http://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionConfiguration.html

mthenw:


function/function.go
```diff
@@ -132,17 +132,10 @@ func (f *Function) Setenv(name, value string) {
 	f.Environment[name] = value
 }
 
-// Deploy code and then configuration.
+// Deploy generates a zip and creates or deploy the function.
+// If the configuration hasn't been changed it will deploy only code,
+// other way it will deploy both configuration and code.
```
tj: "other way" -> "otherwise" :D

tj:
:+1:

PR #224, opened 2016-02-08
fix adding generated files to archive. Closes #221

### LINKED ISSUE #221: Refactor adding bytes to zip
https://github.com/apex/apex/issues/152#issuecomment-180851216

mthenw:
I think this should resolve issue with generated files completely. I reverted changes related to regular files as there is no need to reset mtime for them. 

Only files generated by apex has mtime reseted.

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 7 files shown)


tj:
awesome I'll release that right away

PR #226, opened 2016-02-08
Not ignoring dirs

plukevdh:
Failing test case for .apexignore where directories are not ignored properly.


plukevdh:
Not exactly sure how we want to solve this, but right now, I can't add a `src` dir (or similar) to a function root and then exclude it after building the src into the root function dir.

mthenw:
you can ignore dir with `src/*`

plukevdh:
Hum... that wasn't working before. Appears to now? Closing, thanks.

tj:
hmm `src` should work too if we're matching .gitignore's style, I use that all the time

mthenw:
@tj I'm pretty sure that `src` won't work as we use shell matching patterns

tj:
maybe git does prefix matches or something, because filepath.Match fails to match `src/foo/bar` with `src/*` as well

PR #227, opened 2016-02-08
fix missing profile error when AWS_PROFILE not set

### LINKED ISSUE #223: MissingRegion
Upgraded to  v0.6.0 and started to see an error:

> Error: function versions: MissingRegion: could not find region configuration

~/.aws/config looks like:

```
[default]
region = us-east-1
output = json

[profile x]
region = us-east-1
output = json

```

Everything words fine when I manually set AWS_REGION in env. I'm on Ubuntu 15.04, another user in slack has the same issue on Ubuntu 15.10.

plukevdh:
[fixes #223]

defaults the profile to 'default' per AWS CLI conventions.

### DIFF
--- a/cmd/apex/root/root.go
+++ b/cmd/apex/root/root.go
@@ -84,7 +84,7 @@ func preRun(c *cobra.Command, args []string) error {
 
 	// support region from ~/.aws/config for AWS_PROFILE
 	if profile == "" {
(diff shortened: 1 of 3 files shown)


plukevdh:
New to Go, so any feedback would be :+1:

tj:
thanks!

PR #233, opened 2016-02-10
add cleaning up old versions. Closes #148

### LINKED ISSUE #148: A command to clean-up old versions
Apex makes deploying Go based lambdas so quick that all the binaries start adding up pretty quickly.
It would be nice to be able to clean up old versions or define in project settings that you only want to keep `n` last versions.

mthenw:
Dry run is a bit problematic. It won't show proper info about removing versions :/ @tj do you have any idea how to fix that?

### DIFF
--- a/dryrun/dryrun.go
+++ b/dryrun/dryrun.go
@@ -126,7 +126,12 @@ func (l *Lambda) UpdateFunctionConfiguration(in *lambda.UpdateFunctionConfigurat
 
 // DeleteFunction stub.
 func (l *Lambda) DeleteFunction(in *lambda.DeleteFunctionInput) (*lambda.DeleteFunctionOutput, error) {
-	l.remove("function", *in.FunctionName, nil)
+	if *in.Qualifier == "" {
+		l.remove("function", *in.FunctionName, nil)
+	} else {
+		l.remove("function version", fmt.Sprintf("%s (version: %s)", *in.FunctionName, *in.Qualifier), nil)
+	}
+
 	return nil, nil
 }
 
--- a/function/function.go
+++ b/function/function.go
@@ -67,15 +67,16 @@ func (e *InvokeError) Error() string {
 
 // Config for a Lambda function.
 type Config struct {
-	Description string            `json:"description"`
-	Runtime     string            `json:"runtime" validate:"nonzero"`
-	Memory      int64             `json:"memory" validate:"nonzero"`
-	Timeout     int64             `json:"timeout" validate:"nonzero"`
-	Role        string            `json:"role" validate:"nonzero"`
-	Handler     string            `json:"handler" validate:"nonzero"`
-	Shim        bool              `json:"shim"`
-	Environment map[string]string `json:"environment"`
-	Hooks       hooks.Hooks       `json:"hooks"`
+	Description      string            `json:"description"`
+	Runtime          string            `json:"runtime" validate:"nonzero"`
+	Memory           int64             `json:"memory" validate:"nonzero"`
+	Timeout          int64             `json:"timeout" validate:"nonzero"`
+	Role             string            `json:"role" validate:"nonzero"`
+	Handler          string            `json:"handler" validate:"nonzero"`
+	Shim             bool              `json:"shim"`
+	Environment      map[string]string `json:"environment"`
+	Hooks            hooks.Hooks       `json:"hooks"`
+	RetainedVersions int               `json:"retainedVersions"`
 }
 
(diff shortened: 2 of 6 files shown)


dryrun/dryrun.go
```diff
@@ -126,7 +126,12 @@ func (l *Lambda) UpdateFunctionConfiguration(in *lambda.UpdateFunctionConfigurat
 
 // DeleteFunction stub.
 func (l *Lambda) DeleteFunction(in *lambda.DeleteFunctionInput) (*lambda.DeleteFunctionOutput, error) {
-	l.remove("function", *in.FunctionName, nil)
+	if *in.Qualifier != "" {
```
tj: I'd swap these two branches, `!=` with / `else`

function/function.go
```diff
@@ -489,11 +482,61 @@ func (f *Function) Clean() error {
 	return f.hookClean()
 }
 
+//CleanupVersions removes old function's versions retaing only f.RetainedVersions number
```
tj: `// Cleanup` </ocd> haha

function/function.go
```diff
@@ -265,6 +265,11 @@ func (f *Function) Update(zip []byte) error {
 		"name":    f.FunctionName,
 	}).Info("function updated")
 
+	err = f.CleanupVersions()
```
tj: could be `return f.CleanupVersions()`

tj:
hmm which part of dry runs would be off? relevant stuff should be fine I think, since we're removing the last N versions not the recent N

function/function.go
tobyjoe: :+1: on the cleaner idiom!

function/function.go
```diff
@@ -67,15 +67,16 @@ func (e *InvokeError) Error() string {
...
+	Description      string            `json:"description"`
+	Runtime          string            `json:"runtime" validate:"nonzero"`
+	Memory           int64             `json:"memory" validate:"nonzero"`
+	Timeout          int64             `json:"timeout" validate:"nonzero"`
+	Role             string            `json:"role" validate:"nonzero"`
+	Handler          string            `json:"handler" validate:"nonzero"`
+	Shim             bool              `json:"shim"`
+	Environment      map[string]string `json:"environment"`
+	Hooks            hooks.Hooks       `json:"hooks"`
+	RetainedVersions int               `json:"retainedVersions"`
```
tobyjoe: I wonder if one of those ```.RetainedVersions = -1``` type idioms for "retain 'em all!" is worthwhile? Is that a use case Apex wants to support?

project/project.go
```diff
@@ -74,6 +75,10 @@ func (p *Project) defaults() {
 	if p.NameTemplate == "" {
 		p.NameTemplate = "{{.Project.Name}}_{{.Function.Name}}"
 	}
+
+	if p.RetainedVersions == 0 {
+		p.RetainedVersions = 10
```
tobyjoe: It's probably worth updating deploy.md and functions.md (for consistency) to mention that 10 versions are retained. Or default just to blue/green (ie, ```p.RetainedVersions = 1```)?

function/function.go
tj: probably not, most functions are pretty big once you get a few dependencies going (8MB+) so you'll use the 1.5GB prettttty quick

project/project.go
mthenw: I intentionally put that in one place so there is one source of truth.

I didn't want to set it to 1 because you can specify version while rolling back. Maybe it's too low, maybe too high. Will see :)

mthenw:
> hmm which part of dry runs would be off? relevant stuff should be fine I think, since we're removing the oldest N versions not the recent N

The problem is with `ListVersionsByFunction`. Without `--dry-run` it's called after function update so it returns just deployed version. In dry run mode I can't stub `ListVersionsByFunction` (and push version to results) because it would require calling real `ListVersionsByFunction` :)

function/function.go
tobyjoe: True, true :floppy_disk:

project/project.go
tobyjoe: Got it. Thanks for the response. I'm sure 10 is a good starting point. I always tell me teammates, "When you see a round number, ask about it!" so just trying to be thorough.

tj:
ahhh gotcha hmm

mthenw:
@tj, ok fixed. Not ideal but acceptable (at least by me :))

tj:
woot awesome!

PR #235, opened 2016-02-11
Add go-homedir instead of os/user. Fixes #225 (I think).

### LINKED ISSUE #225: replace os/user with go-homedir
looks like it's still using cgo, cross-compiles fine but seems like it could be our issue with ~/.aws/config on windows. https://github.com/mitchellh/go-homedir

tobyjoe:
This should fix #225 but in full disclosure, I don't have a win
system handy at the moment. That said, no regressions emerge when
running ```go test ./...``` so I'm kinda counting on coverage to
show signs. I can probably test on a win box in the next couple of
days, but welcome others doing so.

Would love :eyes: on this from @tj to make sure the actual issue
was addressed.

Just trying to contribute where I can and this seemed easy...


tj:
LGTM! I don't have access to windows either but this should be fien

PR #236, opened 2016-02-11
fix path separator handling in windows. Closes #222.

### LINKED ISSUE #222: Deploy does not include directories on Windows
Directories in a function's directory are not included when `apex deploy` is run on Windows. So, `require`ing something in `node_modules` will fail, for example.

Deploying the same project on Linux does work. 

In both setups Apex 0.6 is installed.

joelbeckham:
It seems that Lambda does expect / and not \ in the directory paths, so I just force them all to / immediately before adding to zip file. Not the most elegant solution, so if you'd prefer to accomplish this another way, let me know. I tested in both windows and linux and it successfully deploys a working Lambda function.


mthenw:
Thanks!

Issue #229, opened 2016-02-09
`generate` command


pitr:
A simple `apex generate` or `apex init` that will create a simple structure. Takes runtime as argument, defaulting to node. Recreates corresponding [example](https://github.com/apex/apex/blob/master/_examples) in current directory.

(I can work on this)

tj:
:+1: on a related note I'd like to bootstrap Terraform too, like the apex/boilerplate repo (but tweaked a bit). We haven't figured out what works best for node deps yet but we could get started

tobyjoe:
:+1: for sure. 

I spent a couple of nights this week creating a "wrapper" project that automates the creation of an apex/boilerplate style project. It's super rudimentary, and written for my personal workflow, but it may be of interest as reference for ideas. 

I'd happily shift focus to a couple of additional CLI commands. Higher-value ones for myself and my teams would be:

```bash
apex init [path] # Create a new project in the current directory or supplied path
apex new [name] # Create a new function with the given name
```

Additionally, a flag or ```project.json``` and ```function.json``` field to signal that ```terraform apply``` should be called (perhaps via ```deploy``` hook?) would be pretty sweet.

There are obviously lots of opinions on how to manage Terraform config out in the world, but for Apex, it seems like the same "greater specificity" rule as JSON config would be great. 
Either separating the ```terraform``` dir from the ```functions``` dir:

```bash
project.json
functions/foo/index.js
functions/bar/index.js
terraform/project.tf
terraform/modules/iam/iam.tf
terraform/modules/{etc}/{etc}.tf
terraform/functions/foo/function.tf
terraform/functions/bar/function.tf
```

Or, mixing them so each function feels like a Go package:
```bash
project.json
functions/foo/index.js
functions/foo/terraform/function.tf
functions/bar/index.js
functions/bar/terraform/function.tf
terraform/project.tf
terraform/modules/iam/iam.tf
terraform/modules/{etc}/{etc}.tf
```

FWIW, my project hasn't embraced either approach - I'm still just defining one global Terraform for now, to unblock using it on a project at work. My `-h` commands:

```bash
tjbair:test tobyjoe$ copper project -h

Create a new Apex and Terraform project

Usage:
  copper project [flags]

Flags:
  -b, --base string          Project base directory (default ".")
  -s, --bucket string        Terraform remote config bucket
  -c, --credentials string   Credentials (~/.aws/credentials)
  -d, --description string   Project description (default "AWS Lambda function")
[...]

PR #237, opened 2016-02-11
refactor root.preRun to support composition

tobyjoe:
The root.preRun func assigned to ```rootCmd.PersistentPreRunE``` relies
on Project.Open(), which assumes all child commands will want to operate
on an existing Project.

While it's no big deal to assign root.PreRunNoop, that actually seems to
mess with a couple of the truly shared global flag handlers.
For example, ```log-level```, ```chdir``` and ```dry-run``` are handled in the
```root.preRun``` right now. This makes the ```PreRunNoop``` trick come
with the odd side-effect of those flags being unhandled.

The Cobra ```command.SetGlobalNormalizationFunc``` might be abused to handle
flag-triggered actions, but that seems dangerous as well. :skull:

As a starting point, this PR separates out the ```Project.Open()``` and the other
calls, which makes it easier to compose a custom ```PersistentPreRunE``` in
child commands that don't assume a Project already exists.

This is necessary for the work I'm doing today on #229 (which will come in a different PR).

Any better ways of handling this would be awesome. Decomposing the flag-triggered stuff
even further might be good to consider.

### DIFF
--- a/cmd/apex/root/root.go
+++ b/cmd/apex/root/root.go
@@ -74,9 +74,20 @@ func PreRunNoop(c *cobra.Command, args []string) {
 	// TODO: ew... better way to disable in cobra?
 }
 
-// PreRun sets up global tasks used for most commands, some use PreRunNoop
+// preRun sets up global tasks used for most commands, some use PreRunNoop
 // to remove this default behaviour.
 func preRun(c *cobra.Command, args []string) error {
+	err := Prepare(c, args)
+	if err != nil {
+		return err
+	}
+	return Project.Open()
+}
+
+
+// Prepare handles the global CLI flags and shared functionality without
+// the assumption that a Project has already been initialized.
+func Prepare(c *cobra.Command, args []string) error {
 	if l, err := log.ParseLevel(logLevel); err == nil {
 		log.SetLevel(l)
 	}
@@ -119,6 +130,7 @@ func preRun(c *cobra.Command, args []string) error {
 			return err
 		}
 	}
-
-	return Project.Open()
+	return nil
 }
+
+

### CODE REVIEW

**Comment by tj** [comment]
oops might need to rebase, I just pushed a rebase getting rid of some of the merge commits in master

**Comment by tj** [comment]
So far it hasn't been a big concern, the only commands that use the no-op don't really do any AWS related stuff so the log level / dry run not working is ok, but it's definitely not ideal

File: cmd/apex/root/root.go [hunk 1]
```diff
@@ -58,6 +64,7 @@ func init() {
 	f.StringSliceVarP(&Env, "env", "e", nil, "Environment variable")
 	f.StringVarP(&logLevel, "log-level", "l", "info", "Log severity level")
 	f.StringVarP(&profile, "profile", "p", "", "AWS profile to use")
+	f.StringVarP(&creds, "credentials", "c", "", "AWS credentials file to use (~/.aws/credentials)")
```
tj: just realized this probably clashes with our concurrency flag for deploys which is using "c" as well

tobyjoe:
I will resolve the conflicts, @tj - I'm still working on the generator feature, so it's pretty easy. 

On the noOp thing - the global flags not being globally-handled definitely tripped me up - I was assuming, for example, that the dry-run service would be wired up, because I wanted to support the dry-run flag in the #229 generator stuff. But since I can't have ```Project.Open()``` errors kill my command, I had to noOp it, and then there was no dry-run stuff set up... yadda yadda :loop: 

Lemme rebase and I can toss a subordinate PR up in a while that shows a preview of a #229 implementation, so the issue is more clear.

cmd/apex/root/root.go
tobyjoe: Ah, yeah, I'll see if there's a way to detect those clashes in sub-commands. It's tough to have a crowded short-flags space. We could change it to something else, or just not support a short-flag for it?

tobyjoe: I fixed this in #239 by removing the short flag for the ```--credentials``` option. It's a rare-enough option that it shouldn't hurt UX by requiring the longform - though if you have a better short flag, by all means, wire one up. :flags:

tj:
cool looks good to me now! thanks

PR #238, opened 2016-02-11
add IntelliJ files and binaries to the .gitignore

tobyjoe:
To prevent accidental commits (and local git noise) of distribution
binaries and IntelliJ IDEA project/workspace files, I added a few
patterns to the .gitignore file.


mthenw:
I highly recommend adding this to your global .gitignore https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

mthenw:
BTW when developing locally you don't need to build apex. You can just run `go install ./...` and you won't have all those binaries.

tobyjoe:
@mthenw you're totally correct - I need to get my host machine global gitignore working in the little Docker build containers I've been using lately (for a local CI sort of pattern - also why it's running ```make``` instead of ```go install```

I think that makes more sense than adding this noise to the project ```.gitignore``` - I'll kill this one.

mthenw:
:+1:

PR #239, opened 2016-02-11
removes short flag for the AWS credentials file

tobyjoe:
The ```-c``` short flag caused a runtime panic with the deploy command, as it had its own ```-c``` flag. This causes Cobra to :fire: :scream: :fire:

This PR removes the short flag, so the cli now just accepts ```--credentials foo/credzz``` rather than also having a short flag.

### DIFF
--- a/cmd/apex/root/root.go
+++ b/cmd/apex/root/root.go
@@ -64,7 +64,7 @@ func init() {
 	f.StringSliceVarP(&Env, "env", "e", nil, "Environment variable")
 	f.StringVarP(&logLevel, "log-level", "l", "info", "Log severity level")
 	f.StringVarP(&profile, "profile", "p", "", "AWS profile to use")
-	f.StringVarP(&creds, "credentials", "c", "", "AWS credentials file to use (~/.aws/credentials)")
+	f.StringVar(&creds, "credentials", "", "AWS credentials file to use (~/.aws/credentials)")
 
 	Command.SetHelpTemplate("\n" + Command.HelpTemplate())
 }


tobyjoe:
This should actually be using the ```f.StringVar``` vs ```f.StringVarP``` functions of the ```FlagSet``` - updating that now.

tobyjoe:
Ok, switched that function to the better one.

mthenw:
Maybe we should come up with other letter :)

tobyjoe:
@mthenw - I figured it wouldn't hurt UX much to have the long-form, since it should be a rarely-used option, but mostly I didn't want to invent a different flag for it. What's a good pattern? Pick at random? If we change the long form to ```---aws-credentials``` I think ```-a``` is available...

tobyjoe:
I wonder if a pattern like:

- All parent-persistent short flags are capitalized (so, all ```rootCmd``` flags);
- All child command flags are local-only, unless they, too want to have children (none so far in Apex);

Cobra doesn't seem to have any easy way to check at built/test-time.

tj:
I haven't verified but flags before the command might not clash, still a little unintuitive though I suppose

tj:
Ah I thought the VarP variants were for pointers, now I have no clue what the "P" is supposed to be

tobyjoe:
@tj I think the "P" is just because they ran out of letters - like our short flags :100: 

I thought the same thing as you, btw. Instead, it's like, [P=NP](https://en.wikipedia.org/wiki/P_versus_NP_problem) or something.

tobyjoe:
FWIW, I've confirmed that the subordinate flags don't cause panics, so we just need some way for those root short flags to be easier to keep track of. Making them all capitalized reserves a nice 26 item long set for the root, and the sub commands can use lowercase... or vice-versa... I'm one of those "conventions rule" sorta folks, so simple (arbitrary) rules are better than no rules sometimes...

tj:
sounds reasonable to me, the only one I like lower-case is `-e` but we could move that to deploy where it belongs haha

PR #240, opened 2016-02-11
fix loading region from default profile

mthenw:
default sections header is 
```
[default]
```

not 

```
[profile default]
```

 http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -82,7 +82,7 @@ func GetRegion(profile string) (string, error) {
 	}
 
 	sectionName := "default"
-	if profile != "" {
+	if profile != "" && profile != "default" {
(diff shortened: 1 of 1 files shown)


tobyjoe:
Awesome catch. Explains a couple of issues I hit last night... :+1:

tj:
I kinda dislike that they even support defaults haha, just error-prone, even a single profile should be named, oh well lol

tobyjoe:
I don't mind the name "default" but I don't like the "profile" prefix on all non-defaults. It's just odd semantics, and it should be pretty easy to match the default one, given the static name :confused:

PR #241, opened 2016-02-11
fixes shadowed root.Config variable

tobyjoe:
I came across this shadowed ```Config``` variable when trying to inspect ```root.Config```

### DIFF
--- a/cmd/apex/root/root.go
+++ b/cmd/apex/root/root.go
@@ -93,7 +93,7 @@ func Prepare(c *cobra.Command, args []string) error {
 	}
 
 	// credential defaults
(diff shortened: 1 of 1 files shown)


tj:
hmm not at an editor ATM but I'm not even sure we reference Config anywhere haha, maybe we should just remove it

tobyjoe:
Good point, @tj 

It's convenient to have exposed as a bridge for (future?) Terraform integration, but it could just as easily be injected by the ```rootCmd```

tj:
Might as well remove it, if we ever need it we can add it back but having something we're not using makes it confusing

PR #243, opened 2016-02-12
fix pattern matching for .gitignore parity. Closes #228

### LINKED ISSUE #228: fix pattern matching for .gitignore parity
Not reallllly a bug but `src` should be valid to leave out the entire dir etc. re: #226 

Maybe only dirs is what they read specially, I haven't actually read the gitignore docs haha.

mthenw:


tj:
:boo

Issue #250, opened 2016-02-15
Add repl / exec


tj:
This will allow you to poke around in the Lambda container, as well as being useful for testing things out.

It'll be hacky for now since we're coupled with the disk in Function.Open() and a few other spots.

We could just have `apex sh` or `apex sh command`, or both. Open to suggestions for the name, I think shell, exec, or repl would be fine.

tj:
We can also either re-deploy the function each time (and delete on exit), or flag somewhere locally so you don't have to re-build/deploy all the time.

tj:
will live elsehwhere

PR #251, opened 2016-02-15
add start of repl

tj:
re #250 . Need to finish a few things:

- [ ] decide on a name
- [ ] decide if we want to support non-interactive commands as well
- [ ] move to ./repl package
- [ ] change log level to warn
- [ ] add documentation / feature list item

### DIFF
--- a/cmd/apex/main.go
+++ b/cmd/apex/main.go
@@ -15,6 +15,7 @@ import (
 	_ "github.com/apex/apex/cmd/apex/list"
 	_ "github.com/apex/apex/cmd/apex/metrics"
 	_ "github.com/apex/apex/cmd/apex/logs"
+	_ "github.com/apex/apex/cmd/apex/repl"
 	_ "github.com/apex/apex/cmd/apex/docs"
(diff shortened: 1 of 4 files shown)


tj:
screen shot 2016-02-15 at 3 31 40 pm

wolfeidau:
Should the prompt be apex, as apposed to lambda?

What sorts of tasks will the REPL add? Is it just to enable sharing of state between commands?

tj:
Ah it's actually executing commands _in_ a Lambda function, so I think "lambda> " or similar would be good. The primary use-case is to debug AWS roles and inspect the lambda container (check environment variables available etc).

The use-case might not strong enough to have in core. The only way I could see it working for core is if we injected the ability to run commands in _any_ deployed function, so you can inspect local state etc haha, but that's a little weird too. I'd be happy to move it to another repo

mthenw:
I'm also not sure about having it in core :) For starting container purpose `invoke` is fine. Other stuff like debugging/testing can be done inside function IMHO.

tj:
Yea I think I'll move it to apex-shell, thanks guys!

PR #254, opened 2016-02-16
add installation script. Closes #199

### LINKED ISSUE #199: Fancier shell script for binary installs
We have:

```
latest=$(curl -s https://api.github.com/repos/apex/apex/tags | grep name | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)
curl -sL https://github.com/apex/apex/releases/download/$latest/apex_darwin_amd64 -o /usr/local/bin/apex
chmod +x $_
```

Which only works on OSX, but if we improved that a bit and stuffed it in a repo we could have a one-liner that works for everyone. I think this would help improve the initial UX a bit, it really depends how familiar everyone is with the tools I suppose. Regardless I think this would be a handy way to bootstrap an install of _any_ binary release, not just Apex.

mthenw:
I'm not sure about changes in README though.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,21 +5,15 @@ Apex is a small tool for deploying and managing [AWS Lambda](https://aws.amazon.
 
 ## Installation
 
-Download [binaries](https://github.com/apex/apex/releases):
+On OS X or Linux:
 
 ```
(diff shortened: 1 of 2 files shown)


tj:
LGTM!

PR #257, opened 2016-02-18
doc(babel): use webpack to bundle code

mastilver:
using babel and cp node_modules, my lambda is like 45Mb

using webpack with babel, my lambda is 150Kb

### DIFF
--- a/_examples/babel/package.json
+++ b/_examples/babel/package.json
@@ -8,12 +8,13 @@
     "babel-polyfill": "^6.3.14"
   },
   "devDependencies": {
-    "babel-cli": "^6.4.5",
+    "babel-loader": "^6.2.3",
     "babel-plugin-syntax-async-functions": "^6.3.13",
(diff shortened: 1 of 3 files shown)


tj:
would that rollup tree shaking stuff make more sense than webpack? (I haven't tried it yet). Ideally we `npm install --production`, but I guess we could achieve the same thing just by coping modules that aren't dev.

 I'm surprised npm doesn't have a way to _just_ remove dev modules, that might make a cool little Go utility.

tj:
merging either way :D thanks!

mastilver:
> rollup tree shaking stuff make more sense than webpack

webpack@2 will have a similar approch

> Webpack on the server feels a little weird

I felt weird about it too ;)
`target` is set to `node` so it won't do any browser hacks

> webpack's config confuses me

*if* you want to learn more about it: https://github.com/petehunt/webpack-howto

PR #267, opened 2016-02-23
add zeroing of mtime for all files. Closes #262

### LINKED ISSUE #262: Always disregard mtime
Kind of hacky, but since webpack and friends aren't really great at not re-generating the same content, it could be nice to ignore mtime.

tj:


mthenw:
LGTM!

I'm not fan of that but as it's necessary I'm fine :)

Issue #270, opened 2016-02-24
Plugin refactor


tj:
- [ ] no plugin iteration in Function
- [ ] multiple plugins via a `Multi` which replaces what Function has now
- [ ] stateful to help inheritances scenarios where context is lost due to replacing the runtime etc (see babel plugin)

PR #271, opened 2016-02-24
add babel runtime. Closes #217

### LINKED ISSUE #217: Babel runtime
Out of the box Babel support

tj:
Still needs work, lots of TODO's that should be cleaned up, we need a nicer plugin system to facilitate this kind of thing (#270).

first and foremost I think we should refactor some of the plugin logic into utilities so we can use them in other plugins without doing this weird `Runtime` / `Handler` juggling which is a pretty error-prone way to detect other plugins. We badly need stateful plugins so we can recall it between hooks.

I also had to change our nice `AddBytesMtime()` calls since browserify and friends obviously require that these are located on disk. I'm not too concerned about that, not like it introduces real latency but meh. Side-effect of transpiling in node this way is that env won't be usable if you use babel/coffeescript etc without them being officially supported like this.

- [ ] finish TODOs
- [ ] change to webpack or browserify flags to disable polyfills 
- [ ] fix browserify example


tj:
closing for now since it needs work

PR #278, opened 2016-02-28
add zip compression. Closes #263

### LINKED ISSUE #263: Compressed zips

mthenw:
This change works the same on both 1.5 and 1.6. If we would like to change compression level we need to build in 1.6. On 1.5 level param is ignored and `flate.DefaultCompression` is used.


tj:
woot nice

PR #279, opened 2016-02-28
update Readme.md. Closes #265

### LINKED ISSUE #265: update README
There is some old info. Also "Example" section could be removed and reference to website should be added.

mthenw:
- added description from website
- added link to examples below runtime list for easier access
- removed Golang example as it's the same as in go-apex and in _example dir
- removed Contributors section as it's the same as https://github.com/apex/apex/graphs/contributors
- added link to Twitter

@tj if you don't like any of those just let me know :)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,7 +1,7 @@
 
 ![Apex Serverless Architecture](assets/logo.png)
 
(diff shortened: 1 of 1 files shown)


tj:
lgtm, I think maybe the contributors section should just be people over N commits or something, aka maintainers, but doesn't really matter to me

PR #290, opened 2016-03-01
add infra command

mthenw:
I will do that in small steps. This is bare `infra` command extracted from `add-init` branch. As there is no TF -> apex integration yet I didn't include env switcher.

### DIFF
--- a/cmd/apex/infra/infra.go
+++ b/cmd/apex/infra/infra.go
@@ -1,0 +1,44 @@
+// Package infra proxies Terraform commands.
+package infra
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/apex/apex/cmd/apex/root"
+	"github.com/apex/apex/infra"
+)
+
+var infraEnvironment string
+
+// example output.
+const example = `  View change plan
+  $ apex infra plan
+
+  Apply changes
+  $ apex infra apply`
+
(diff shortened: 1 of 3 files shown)


mthenw:
@tj review plox

tj:
LGTM but I think we still need the notion of an env, but works for now!

cmd/apex/infra/infra.go
```diff
@@ -0,0 +1,44 @@
...
+package infra
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/apex/apex/cmd/apex/root"
+	"github.com/apex/apex/infra"
+)
+
+var infraEnvironment string
```
tj: think this is dead code

mthenw: damn!

mthenw:
> LGTM but I think we still need the notion of an env, but works for now!

yeah but can we support env for infra but not for functions?

tj:
I think we'll need it globally, for example to swap env variables for staging and prod, that's something I'll need in every app so I think a global notion would be good

mthenw:
absolutely, I was asking more if we can have env flag only for infra in current moment.

tj:
ohhh yea for sure

PR #294, opened 2016-03-02
add global flag to prevent colorizing output. Closes #293

### LINKED ISSUE #293: No colorization / formatting for better interop with other programs
I've actually already written a PR for this but saw that I should open an issue first. Whoops.

At any rate, it would be handy to have a `--no-color` flag for the apex command that prevents colorizing output so you can pipe it to other programs. `apex list` is useful to get function names and then pipe it through an awkward `sed` to strip out the colors. This flag would help.

We could also use something like isatty to format the output differently for pipes but I think that would constitute a separate issue.

ajcrites:
Colorization is only done in a couple spots right now, but ultimately it would be useful to have something like a `Printer` class for output with methods like `.Blue`, `.Red`, etc. If this flag is set, the `Printer` could be an implementation that doesn't actually colorize the output for these methods.

### DIFF
--- a/cmd/apex/list/list.go
+++ b/cmd/apex/list/list.go
@@ -69,7 +69,11 @@ func outputTFvars() {
 func outputList() {
 	fmt.Println()
 	for _, fn := range root.Project.Functions {
-		fmt.Printf("  \033[%dm%s\033[0m\n", colors.Blue, fn.Name)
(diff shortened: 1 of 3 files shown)


tj:
closing for now, we'll get a complete version in some time

PR #296, opened 2016-03-02
add creating TF configs by init command

mthenw:
@tj, it's a bit different that you code. Let me know what do you think.

TODO:
- [ ] don't validate role if TF infra

There is still no env switch support. I will make another PR for that.


mthenw:
Still not sure if using `terraform output` is fine.

boot/boot.go
```diff
@@ -4,47 +4,84 @@ import (
...
 	help(`Enter the name of your project. It should be machine-friendly, as this is used to prefix your functions in Lambda.`)
 	name := prompt.StringRequired("  Project name: ")
 
 	help(`Enter an optional description of your project.`)
 	description := prompt.String("  Project description: ")
 
-	help(`Enter your AWS Account ID.`)
-	accountID := prompt.StringRequired("  AWS Account ID: ")
 	fmt.Println()
+	if prompt.Confirm("Would you like to manage infrastructure with Terraform? (yes/no)? ") {
```
tj: two `?`'s

boot/boot.go
```diff
@@ -4,47 +4,84 @@ import (
...
+		fmt.Println()
+		if err := initProject(name, description, ""); err != nil {
+			return err
+		}
+
+		if err := initInfra(); err != nil {
+			return err
+		}
+
+		help("Setup complete!\nNext steps: \n  - apex infra apply - apply Terraform configs\n  - apex deploy - deploy example function")
```
tj: probably good to have `apex infra plan` in there too, might help illustrate that it just delegates to terraform

tj:
LGTM at a glance. Shelling out to terraform for the role each time could probably be replaced by reading from `.terraform/terraform.tfstate`

mthenw:
Is it stored locally when using remote state?

tj:
Yep it's still cached locally

tj:
screen shot 2016-03-02 at 2 25 16 pm

mthenw:
well it's easier exec `terraform output` :) and it is done only once per `apex ...` call

tj:
true true, if it becomes noticeably slow we can change

PR #297, opened 2016-03-02
add support for role from infrastructure

mthenw:


function/function.go
```diff
@@ -132,6 +133,10 @@ func (f *Function) Open() error {
 		return err
 	}
 
+	if f.Config.Role == "" && !f.InfraExists {
+		return errors.New("role is required")
```
tj: hmm shouldn't we error even if you have ./infrastructure and it's still empty?

mthenw: Yeah. The problem is with initial `apex infra plan/apply`. We load project (which means those checks will fire) but there still no infra applied.

mthenw: No idea how to cope with that.

PR #312, opened 2016-03-08
add deploy, rollback, and invoke --alias support. Closes #7

### LINKED ISSUE #7: Aliases
Provide commands for manually working with aliases. We don't recommend using aliases for environments, however they may still be useful in some cases, such as per-person aliases in a staging environment.

- [x] deploys
- [x] rollback
- [x] invoke
- [x] test
- [x] update tests so they pass
- [x] add tests
- [ ] maybe default flags to CurrentAlias so people can see that

tj:
Review please!

Adds `--alias name` to rollback, deploy, and invoke so that you can launch canary functions etc.

I'll squash that other lame commit after.

### DIFF
--- a/cmd/apex/deploy/deploy.go
+++ b/cmd/apex/deploy/deploy.go
@@ -15,12 +15,18 @@ var env []string
 // concurrency of deploys.
 var concurrency int
 
+// alias.
+var alias string
+
 // example output.
 const example = `  Deploy all functions
   $ apex deploy
 
   Deploy specific functions
   $ apex deploy foo bar
+
+  Deploy canary alias
+  $ apex deploy foo --alias canary
 
   Deploy functions in a different project
   $ apex deploy -C ~/dev/myapp`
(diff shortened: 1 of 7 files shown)


mthenw:
What about `delete`? I would assume that `apex delete -a sth` would delete that alias.

cmd/apex/deploy/deploy.go
```diff
@@ -39,12 +45,15 @@ func init() {
 
 	f := Command.Flags()
 	f.StringSliceVarP(&env, "env", "e", nil, "Environment variable")
+	f.StringVarP(&alias, "alias", "a", "", "Function alias")
```
mthenw: I think it's a good idea to pass "current" as a default here.

tj:
Hmm yea regarding delete I'm not sure if I would expect that to remove the alias or to remove the version specified by the alias. I think since multiple aliases could be pointing at the same version it might be best to leave that out for now

tj:
squashed and added the default!

PR #313, opened 2016-03-08
fix detecting config changes. Closes #311

### LINKED ISSUE #311: Config always detected as changed
At least that's what I'm seeing now. We might be better off just json-marshalling that thing and doing a string comparison

mthenw:
@tj not sure if this is what you mean but it works :)

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -12,7 +12,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


function/function.go
```diff
@@ -576,39 +587,40 @@ func (f *Function) currentVersionAlias() (*lambda.AliasConfiguration, error) {
...
-	}
+	remoteConfig, _ := json.Marshal(diffConfig{
+		Description: *config.Configuration.Description,
+		Memory:      *config.Configuration.MemorySize,
+		Timeout:     *config.Configuration.Timeout,
+		Role:        *config.Configuration.Role,
+		Handler:     *config.Configuration.Handler,
+		VPC: vpc.VPC{
+			Subnets:        aws.StringValueSlice(config.Configuration.VpcConfig.SubnetIds),
+			SecurityGroups: aws.StringValueSlice(config.Configuration.VpcConfig.SecurityGroupIds),
```
tj: the rest looks good to me but what about the case when `config.Configuration.VpcConfig` == nil

mthenw: won't be. it's initialized with empty arrays.

mthenw:
I'm merging as the rest is fine :)

function/function.go
tj: I don't think it's initialized on the AWS end properly, I was having errors but I'll install master's HEAD and check in a bit

mthenw: yeah, please check. based on my tests it's initialized.

tj: weiiird seems fine, not sure what I hit

PR #327, opened 2016-03-11
Added "GitHub API rate limit exceeded" error msg

fibo:


install.sh
```diff
@@ -42,9 +42,13 @@ fi
 LATEST=$(curl -s https://api.github.com/repos/apex/apex/tags | grep name | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)
 URL="https://github.com/apex/apex/releases/download/$LATEST/apex_$PLATFORM"
 
-curl -sL https://github.com/apex/apex/releases/download/$LATEST/apex_$PLATFORM -o /usr/local/bin/apex
-chmod +x $_
-
+if [ -z $LATEST ] ; then
+  echo "Sorry, GitHub API rate limit exceeded. Download binary from https://github.com/apex/apex/releases"
```
tj: Not sure we should assume that a failure means it's the API limit, maybe we should just say `Error requesting. Download binary ...`, not sure, unless the request's error output is decent we could use that

fibo:
No prob, I updated the error message suggested.

tj:
awesome thanks!

Issue #300, opened 2016-03-03
Pre-built Java JAR support


robzienert:
The current Java plugin requires Apex to have a POM file and then rebuilds the JAR for you. A lot of more modern JVM projects use Gradle, and generating a POM from Gradle is a pain, and is redundant. 

It would be nice if the java plugin were made to support either Gradle, or supplying a fat JAR. What I could see is one of the two:

1. Add a fork in Build to look for a `build.gradle`, in which case it would attempt to run the `gradle shadowJar distZip` tasks. If `build.gradle` doesn't exist, it would move onto generating the pom file as it does today.
2. It could just look for the presence of `target/apex-plugin-target.jar` or `build/apex-plugin-target.jar`. If those exist, it'll zip the JAR and move on. 

Maybe just making a `java-gradle` plugin would be the cleaner way?

Thoughts? This work would be pretty easy, but would remove a lot of boilerplate legwork for projects using Gradle.

tj:
Whatever is idiomatic sounds good to me, I'm a total Java noob so we're relying on contributions for the integration. For anything more opinionated I like the idea of "sub-runtimes" like node-babel or like you mention java-gradle.

bhurlow:
supplying a built jar would help for clojure support as well

mthenw:
Would be great to have Gradle support. I think that there should be two separate plugins as those are two different systems.

I think that option 1. is how we should approach that. But not in existing plugin but in separate.

robzienert:
Separate plugins sounds good to me. Nice and clean. I'll work on a `java-jar` plugin, since that's quite ubiquitous.

mthenw:
Also I think that we should stop generating pom. It adds unnecessary noise in plugin code and probably isn't helpful to anyone. 

As it hasn't been released yet I will refactor it tomorrow and also add Gradle support.

mthenw:
@robzienert haven't seen your. Ok so I will refactor existing plugin and you can create new one for Gradle. Not sure why 'java-jar' not 'java-gradle'?

robzienert:
Based on @bhurlow 's comment, it may make sense to have a generic jar plugin. The JVM has Maven, Gradle, SBT, Lein and more - all that do builds differently, but at the end of the day are all capable of generating a single fat jar.

[...]

PR #330, opened 2016-03-13
Adding java-jar plugin

robzienert:
Adds a `javajar` plugin, as discussed in #300. 

Searches common build artifact paths for a jar name by `apex.jar`. I think I might change the findJar function to find by JAR classifier instead of forcing the JAR to be a static name. Example: `apex.jar` vs `artifact-1.0.0-SNAPSHOT-apex.jar`. Thoughts?

### DIFF
--- a/plugins/java/java.go
+++ b/plugins/java/java.go
@@ -3,13 +3,11 @@ package java
 
 import (
 	azip "archive/zip"
-
+	"errors"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
-
-	"errors"
+	"strings"
 
 	"github.com/apex/apex/function"
 	"github.com/jpillora/archive"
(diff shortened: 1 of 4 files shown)


tj:
LGTM with my limited java knowledge! I'll let @mthenw take a peek too

plugins/javajar/javajar.go
```diff
@@ -0,0 +1,107 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/apex/apex/function"
+	"github.com/jpillora/archive"
+)
+
+const (
+	// Runtime name used by Apex
+	Runtime = "javajar"
```
tj: wonder if we should do "java-jar" and "node-babel" etc with the others in the future instead of "babel" etc

robzienert: Happy to make that change.

mthenw:
Could you also update https://github.com/apex/apex/blob/master/docs/java.md?

mthenw:
Also I think it's worth to use https://github.com/apex/apex/blob/master/plugins/inference/inference.go#L21.

robzienert:
I explicitly left it out of inference since it can search multiple directories. If we go the path of not using a static filename, it becomes even more difficult to infer as well.

mthenw:
I'm wondering if we need separate plugin for java. Maybe java-jar is enough and we should remove previous Java support. @tj what do you think?

tj:
I'm impartial, I don't know enough about the ecosystem to comment on which idioms make sense by default.

plugins/javajar/javajar.go
```diff
@@ -0,0 +1,107 @@
...
+			return err
+		}
+
+		b, err := ioutil.ReadAll(r)
+		if err != nil {
+			return err
+		}
+		r.Close()
+
+		zip.AddBytes(file.Name, b)
```
davidporter-id-au: @robzienert Correct me if I'm wildly off here, but do you want to zip the already compressed `.jar` file?

robzienert: This isn't immediately obvious, but a plugin is given a zip archive that it needs to add files to. The jar itself is a zip archive, so I'm just iterating the files inside of the jar and adding them to the provided zip. It's effectively the same as renaming `foo.jar` to `foo.zip` while using this plugin.

robzienert:
Just got back from vacation, I'll work on changing this PR to rework the standard `java` plugin, if that's what we're arrived at, @tj @mthenw.

robzienert:
Replaced the Java plugin with java-jar.

tj:
I really need to learn at least enough Java to be useful here hahaha. LGTM though ;p – @mthenw?

plugins/java/java.go
```diff
@@ -20,51 +18,58 @@ const (
...
-	function.RegisterPlugin("java", &Plugin{})
+	function.RegisterPlugin(Runtime, &Plugin{})
 }
 
-// Plugin implementation.
+// Plugin implementation
 type Plugin struct{}
 
-// Open adds java defaults.
+// Open adds java-jar defaults. No clean operation is implemented, as it is
```
mthenw: no "java-jar"

mthenw:
LGTM

Could you only fix comment for Open func?

mthenw:
Actually. I will fix that after merge. Thanks!

PR #331, opened 2016-03-14
add env switch. Closes #304

### LINKED ISSUE #304: Add env switching
Based on https://github.com/apex/apex/issues/253

mthenw:
I will add setting up other envs in separate PR :)

### DIFF
--- a/boot/boot.go
+++ b/boot/boot.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/apex/apex/boot/boilerplate"
@@ -40,7 +41,7 @@ var remoteStateCommand = `
     -backend=s3 \
     -backend-config="region=%s" \
     -backend-config="bucket=%s" \
-    -backend-config="key=terraform/state"
+    -backend-config="key=terraform/state/%s"
 `
 
 // All bootstraps a project.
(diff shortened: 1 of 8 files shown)


mthenw:
@tj could you take a look?

tj:
nice LGTM, maybe add the default to the flag as well. I wonder if we should use "stage" as well, instead of dev for default

mthenw:
@tj or maybe there should be a question:

`Which environment do you want to setup (dev, test, stage, prod, none)?`

and until `none` it will be asking in a loop.

tj:
hmm yea that sounds like a good plan. Maybe:

- `List the environments you would like: stage,prod` (something like that)
- `Which environment should be the default? stage`

maybe we store that in ./project.json?

mthenw:
@tj updated

boot/boot.go
```diff
@@ -109,31 +114,74 @@ func initProject(name, description, iamRole string) error {
...
+
+// setupEnv creates environment dir
+func setupEnv(env string) error {
+	logf("creating %s environment", env)
+
+	if err := os.MkdirAll(filepath.Join(infra.Dir, env), 0755); err != nil {
+		return err
+	}
+
+	maintf := filepath.Join(infra.Dir, "_env", "main.tf")
```
tj: what's this ./infrastructure/_env/main.tf for? (might be reading that wrong)

tj:
cool looks good! Just wondering about that one _env dir

mthenw:
What about "_env"? You want me to rename it?

boot/boot.go
mthenw: it's just a placeholder :)

PR #332, opened 2016-03-15
add java env variables. Closes #272

### LINKED ISSUE #272: Support env vars in Java

skarnecki:
- not sure if cleaning properties is required
- assumed loading properties file should be on user-side


_examples/java/functions/with-pom/src/main/java/lambda/Example.java
```diff
@@ -43,6 +58,38 @@ public ExampleResponse() {
...
+        String token = "";
+        try {
+            token = (String) getProperty("LOGGLY_TOKEN");
+        } catch (IOException e) {
+            System.out.println(e.toString());
+        }
+        return new ExampleResponse(event.getHello(), token);
+    }
+
+    private Object getProperty(String key) throws IOException {
```
mthenw: is there a way to achieve the same without code changes?

skarnecki: As discussed @Slack: there are couple ways of getting .properties in Java. I've used most generic one in example. It should be up to user

mthenw:
Closing as the Java plugin changed.

PR #341, opened 2016-03-22
add passing aws_region var to Terraform

mthenw:
It's needed for integrating with API Gateway, but it also might be useful for other integrations.


tj:
lgtm! not sure we need the "Generated by X" thing just because most people would be editing those files at some point down the road

mthenw:
@tj removed :) thanks.

Issue #37, opened 2016-01-05
API Gateway source


mthenw:
I've opened new issue as the previous one is more about multi-function projects. This is the next task that I would take care of.

Have you thought how should be function source (this or any other like S3 event) defined in lambda.json?

tj:
Haven't given that stuff too much thought yet since we'll need a way to let the dev override configuration nicely. Started #30 with some ideas, but if I have time tonight I'll probably start on the multi-function stuff, getting rid of chdir etc.

mthenw:
One question is if we want to create all needed resources or we should reuse existing. In terms of API Gateway it's obvious that those resources are tightly coupled with Lambda functions. I'm not so sure about that in terms of S3 though. But this is different topic :)

Assuming that we want to create API Gateway stuff I would expect that at least I can:

* define HTTP method per function
* define url path per function 
* toggle CORS per function/project
* define Throttling Settings per project
* toggle CloudWatch logs/metrics per project
* toggle API cache per project 

Workflow (deploy):

* pre project deploy (check if API exists, if not create)
* post function deploy (check if Resources/Route and Method exists, if not create)

Workflow (delete):

* post function delete (delete not used Route and Method)

johannesboyne:
How about a JSON configuration inside the `project.json` or somewhere else with the following - definitely t.b.d. structure: https://gist.github.com/johannesboyne/4a85bb53842cd4bf8715

Direction OK?

tj:
Related to #59, but I think we need to think about some more. Especially the environment config and sources/api gateway so we get it right. I haven't put much thought into it yet since I'm not there in my project(s) but I'll try and do so tonight and get some more ideas up. I'm cool with JSON, but I want to make sure the shape is DSL-friendly for later and generally not leaky. I'll check your stuff out tonight too!

tj:
need to speak with the Gateway PM again, see where they're at

zanona:
By any chance, have you guys heard about [Swagger](http://swagger.io)?
API Gateway uses this format to export stages where it’s also possible to import the whole API structure by using [awslabs/aws-apigateway-swagger-importer](https://github.com/awslabs/aws-apigateway-swagger-importer). — That seems a good way towards a standard.

tj:
@zanona yep! We have a near-term plan for integration

plukevdh:
[...]

PR #342, opened 2016-03-22
add API Gateway integration example. Closes #339

### LINKED ISSUE #339: API Gateway example via TF
https://github.com/hashicorp/terraform/pull/4295#issue-121938495

mthenw:
Next step could be creating module to simplify that...


tj:
haha nice! yea definitely could be simpler

amccloud:
@mthenw @tj Should apex or another tool handle api gateway orchestration? Would it be okay to generate a terraform config from a `routes` key to the project.json (or routes.json file).

```json
{
  "name": "with-gateway",
  "routes": {
    "/examples/{exampleId}": {
      "method": "GET",
      "function": "hello"
    }
  }
}
```

mthenw:
@amccloud we are tracking API Gateway support in https://github.com/apex/apex/issues/37. It's pretty complicated right now and we are not sure if we want invest time in that.

tj:
Ditto, in a lot of ways I think the approach AWS took with API Gateway is not a very good one. Writing an API is easy, we just need a way to get it in the Lambda containers, they complicated the wrong thing. I wouldn't mind supporting the swagger stuff just because the import API is so easy now, but I still think API Gateway is a weird solution

PR #345, opened 2016-03-27
Rename --start to --since. Closes #326

### LINKED ISSUE #326: Rename --start to --since
This is what Docker uses, they both read sort of weird with durations but I think it's a tiny bit better

farazfazli:
#326


tj:
thankya!

PR #347, opened 2016-03-27
fix: api-gateway example Internal server error

amowu:
When API Gateway method response status code is 200, and has not define model, it will be response 500 error like:

  Execution failed due to configuration error: Output mapping refers to an invalid method response: 200
  Method completed with status: 500

So I added the default model "Empty" to resolve this issue.


tj:
thanks!

PR #357, opened 2016-04-07
add nodejs 4.3.2 support. Closes #356

### LINKED ISSUE #356: Add support for node 4.3.2 runtime
Amazon recently added a nodejs 4.3.2 runtime for lambda. It would be able to specify it as a value in the configuration json files.  Maybe like:

```
{
  "description": "<function desc> ",
  "runtime": "nodejs:4.3.2"
}
```

mthenw:
Worth to mention that updating nodejs version on already deployed function will have no effect. `apex delete` is needed before.

### DIFF
--- a/plugins/inference/inference.go
+++ b/plugins/inference/inference.go
@@ -16,7 +16,7 @@ func init() {
 	function.RegisterPlugin("inference", &Plugin{
 		Files: map[string]string{
 			"main.py":             python.Runtime,
-			"index.js":            nodejs.Runtime,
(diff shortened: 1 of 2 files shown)


mthenw:
@tj I'm not sure about inference. Should we switch to 4.3? Hmm, I will add it.

tj:
Hmm good question haha, I suppose if it's unspecified at both levels assuming the recent is fine

tj:
If we need later we could have a `RuntimeLatest = RuntimeWhatever`

PR #365, opened 2016-04-14
Update go runtime to use node 4.3

dcu:
I already tested it on AWS and it works great.
Let me know if there's something else I need to do to get this approved.


mthenw:
Not sure about that but based on https://aws.amazon.com/blogs/compute/node-js-4-3-2-runtime-now-available-on-lambda/ (shim)[https://github.com/apex/apex/blob/master/shim/index.js] needs to be updated also.

dcu:
@mthenw I updated to use the new callback argument however it was optional and actually I had to set callbackWaitsForEmptyEventLoop to false otherwise it'd timeout.

it says it's optional here: http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html#nodejs-prog-model-handler-callback

let me know if I should revert this change.

mthenw:
Oh sorry, I didn't read to the end. If it's working without that you can revert last commit.

dcu:
I tested both(with and without the new callback) and I didn't notice any difference

mthenw:
Let's leave it as it is. Thanks.

tj:
I have some stuff I can use to test and see if there's a perf benefit to upgrading or not at some point

simshanith:
I think 8a270e84 does everything necessary. I manually update my apex go deployment to use Nodejs 4.3 via the AWS Console, and everything works as expected.

b9be3f765965 reverted this PR #365 ; I think if 8a270e84 is reapplied it should work fine without the changes in e59845fe2a0d1c2584df96292dda72cb2079f79

tj:
I'll have to give it a try with something non-production haha, I don't see a compelling reason to change it really since the shim doesn't do much

PR #366, opened 2016-04-14
Adding example of using gopherjs

raff:
This is an example of a nodejs Lambda function where the code is written in Go and converted to Javascript using gopherjs. It requires "gopherjs" to be installed (https://github.com/gopherjs/gopherjs).

The generated code is a little big, especially when using Go standard packages, but it may be useful when you have so Go code you want to use but you don't want to use the proxy/skim.

It would also be interesting to add "gopherjs" as a specific runtime and possibly expose via apex a gopherjs library that simplify interacting with the native nodejs runtime and/or AWS.


tj:
hmm haha I'm not sure how robust gopherjs is, if we'd want to advocate using it or not, you haven't had issues?

mthenw:
IMHO it might be confusing to add this example if we support golang ~natively.

PR #367, opened 2016-04-16
add human-friendly duration output for metrics

tj:
"duration: 24 minutes "  etc instead of "123124123213ms" haha, rounding is not good but seeing a huge number is not very helpful. Ideally we have "5 hours and 36 minutes" or something


tj:
nvm I don't like the rounding

PR #368, opened 2016-04-17
Add metric price

tj:
adds Lambda's cost output to `metrics`, pretty handy

screen shot 2016-04-16 at 5 00 57 pm


mthenw:
Wow that's awesome.

mthenw:
Not sure about that but aren't those prices only for US regions? What about other regions?

tj:
hmm good question, I don't see any per-region pricing, might be ok

tj:
<3 that squash button

PR #375, opened 2016-04-20
add support for symlinks. Closes #320

### LINKED ISSUE #320: Support symlinks
I put all the shared code for my functions in a shared npm module. To make development easier, I use "npm link" to create a symbolic link of this module in each function. This way, as I make changes to my shared code, I don't have to go and do an "npm install" for each function.

However, this approach appears to not work, because when I ran "apex deploy" I got this error:
Error: function getTasks: Only regular files supported: node_modules/core.

What's the recommended way to handle shared code for the Node.js environment?

mthenw:
Not sure if we need any docs update or example?


tj:
LGTM 💥. An example and maybe a quick mention in the docs for the use-cases for shared node/py code would be pretty rad but we can add that after

PR #379, opened 2016-04-22
Handle panic when trying to set incomplete -s var

### LINKED ISSUE #350: Environment variables with commas cause `index out of range`
This may be related to another library apex is using, but I can't tell. If I try to set an environment variable that contains a comma, I get an "index out of range" error, e.g.

    apex deploy --set "foo=bar,baz"

Using `'foo="bar,baz"'` or `"foo='bar,baz'"` doesn't seem to change anything.

I'm not trying to use an array or set multiple variables or anything like that; the environment variable I'm trying to set is a MongoDB DSN with a replica set so it happens to have a comma in it.

```
panic: runtime error: index out of range

goroutine 1 [running]:
panic(0xa7ae80, 0xc82000e220)
        /usr/local/go/src/runtime/panic.go:464 +0x3e6
github.com/apex/apex/cmd/apex/deploy.run(0xed1da0, 0xc8201778c0, 0x0, 0x12, 0x0, 0x0)
        /Users/tj/dev/src/github.com/apex/apex/cmd/apex/deploy/deploy.go:65 +0x282
github.com/spf13/cobra.(*Command).execute(0xed1da0, 0xc8201777a0, 0x12, 0x12, 0x0, 0x0)
        /Users/tj/dev/src/github.com/spf13/cobra/command.go:565 +0x62c
github.com/spf13/cobra.(*Command).ExecuteC(0xed2fa0, 0xed1da0, 0x0, 0x0)
        /Users/tj/dev/src/github.com/spf13/cobra/command.go:656 +0x55c
github.com/spf13/cobra.(*Command).Execute(0xed2fa0, 0x0, 0x0)
        /Users/tj/dev/src/github.com/spf13/cobra/command.go:615 +0x2d
main.main()
        /Users/tj/dev/src/github.com/apex/apex/cmd/apex/main.go:37 +0x81
```

eyko:
Environment variables passed via --set should be in the key=value format. Calling deploy can cause a panic in any of the following scenarios:
```shell
apex deploy -s FOO
apex deploy -s FOO=
apex deploy -s FOO=bar,BAZ
... etc
```
This commit adds an error message and returns an error, rather than an unfriendly `panic` with no explanation (other than the error causing the panic). As such, this doesn't change current behaviour and the command will still exit. 

Another way this can be handled is by setting those environment variables to an empty string (`root.Project.Setenv(parts[0], "")` and neither panic nor exit. Question is... does the user ever intend to set an empty environment variable?

Closes #350


cmd/apex/deploy/deploy.go
```diff
@@ -61,8 +63,13 @@ func run(c *cobra.Command, args []string) error {
...
 
 	for _, s := range env {
-		parts := strings.Split(s, "=")
-		root.Project.Setenv(parts[0], parts[1])
+		parts := strings.SplitN(s, "=", 2)
+		if len(parts) == 2 {
+			root.Project.Setenv(parts[0], parts[1])
+		} else {
+			log.Fatalf(`Environment variable "%s" has no value.`, parts[0])
+			return errors.New("environment variable needs a value")
```
tj: maybe just a fmt.Errorf here with the var name?

tj:
LGTM other than maybe the Errorf thing, I agree that an error there is probably the way to go

cmd/apex/deploy/deploy.go
eyko: Do you mean, just return the error and no need for log.Fatalf? as in:  
```go
} else {
        return errors.New(fmt.Errorf("environment variable %s needs a value", parts[0]))
```
?

tj:
:boom

eyko:
Hold on that was incomplete- probably introduced a bug there. Should be returning `fmt.Errorf` and cleaning up deps after that (log / errors respectively)

eyko:
Oh never mind, I see you already fixed that in dfc89e06b4a7f5f87f0956b32daed5d280cf4f58 👍

PR #388, opened 2016-05-01
refactor boilerplate to use callback param

tlvince:
See:
http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-using-old-runtime.html#transition-to-new-nodejs-runtime


mthenw:
👍

Issue #392, opened 2016-05-03
Log tailing rate limit


tj:
If you have quite a few funcs and you try to tail them all it's pretty easy to hit the rate limit, not realllly a bug but marking as one for now. We could probably just detect and backoff, or allow the user to set the interval at least.

Less of a problem once we have groups or similar, but sometimes I like to `apex logs -F level=error -f` to make sure I didn't botch something

```
ThrottlingException: Rate exceeded
	status code: 400, request id: 43c79cf5-1150-11e6-b87f-39da7ea44302
```

tj:
mitigated now with larger interval

PR #394, opened 2016-05-04
Fix logs

tj:
@mthenw 

looks like I botched this before haha, forgot to paginate. reduced PollInterval as well but that one is probably better as its own commit so no need to squash, might help with #392


tj:
63578272830812135630833659_1435176367_annoyed-obama

PR #396, opened 2016-05-06
add basic analytics

tj:
Basic stuff for now.

Later it would be nice to track errors / panics, can't think of anything else for now.

It'll flush every 500 events, or once every 24 hours, introducing ~300ms of latency for that time.


mthenw:
💣

PR #399, opened 2016-05-06
Fixes #398 - Allow relative symlink values during deploy

### LINKED ISSUE #398: Relative symlinks not properly followed in deployment
It seems there's an issue with evaluating relative symlinks during deployment.

Changing from ```os.ReadLink``` to ```filepath.EvalSymlinks``` seems to fix it, with a couple of minor changes in related code. 

Submitting a PR to discuss, but I've tested it locally and it seems good...

tobyjoe:
The implementation before this PR leads to errors opening symlink'ed files that happen to be relatively-linked. This change fixes that issue, adds a couple of helpful debug calls, and unmasks the ```f``` variable in the function body.

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -542,34 +542,37 @@ func (f *Function) Build() (io.Reader, error) {
 
 		info, err := os.Lstat(filepath.Join(f.Path, path))
 		if err != nil {
+			f.Log.Debugf("Lstat error: %v", err)
 			return nil, err
 		}
 
-		realPath := path
+		realPath := filepath.Join(f.Path, path)
 		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
-			linkPath, err := os.Readlink(filepath.Join(f.Path, path))
+			linkPath, err := filepath.EvalSymlinks(filepath.Join(f.Path, path))
 			if err != nil {
+				f.Log.Debugf("EvalSymlinks error: %v", err)
 				return nil, err
 			}
 			realPath = linkPath
 		}
 
-		f, err := os.Open(filepath.Join(f.Path, realPath))
+		fh, err := os.Open(realPath)
 		if err != nil {
+			f.Log.Debugf("Open error: %v", err)
 			return nil, err
 		}
 
-		info, err = f.Stat()
+		info, err = fh.Stat()
 		if err != nil {
 			return nil, err
 		}
 
 		unixPath := strings.Replace(path, "\\", "/", -1)
-		if err := zip.AddInfoFile(unixPath, timelessInfo{info}, f); err != nil {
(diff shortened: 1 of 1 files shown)


mthenw:
Could you provide case in which current solution doesn't work? I've just checked it and work fine for me.

tobyjoe:
@mthenw - the project I hit it on is internal & not OSS, but I can make a test project.

The long-story-short was that a nodejs dependency had a ```bin``` command file that was relatively-linked during ```npm install``` and the path joining logic before this PR didn't compute the relative value properly relative to the function root. 

Gimme a few to make a test project and toss it up on :octocat:

mthenw:
Thanks!

tobyjoe:
Here ya go, @mthenw - https://github.com/tobyjoe/apex-issues/tree/master/functions/issue-398

The README shows how I tested it...

mthenw:
Thanks again, make sense! Could you only remove those who log calls, we don't want to drown in log lines :)

function/function.go
```diff
@@ -542,34 +542,37 @@ func (f *Function) Build() (io.Reader, error) {
 
 		info, err := os.Lstat(filepath.Join(f.Path, path))
 		if err != nil {
+			f.Log.Debugf("Lstat error: %v", err)
```
mthenw: this

function/function.go
```diff
@@ -542,34 +542,37 @@ func (f *Function) Build() (io.Reader, error) {
...
 			return nil, err
 		}
 
-		realPath := path
+		realPath := filepath.Join(f.Path, path)
 		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
-			linkPath, err := os.Readlink(filepath.Join(f.Path, path))
+			linkPath, err := filepath.EvalSymlinks(filepath.Join(f.Path, path))
 			if err != nil {
+				f.Log.Debugf("EvalSymlinks error: %v", err)
```
mthenw: this :)

function/function.go
```diff
@@ -542,34 +542,37 @@ func (f *Function) Build() (io.Reader, error) {
...
 				return nil, err
 			}
 			realPath = linkPath
 		}
 
-		f, err := os.Open(filepath.Join(f.Path, realPath))
+		fh, err := os.Open(realPath)
 		if err != nil {
+			f.Log.Debugf("Open error: %v", err)
```
mthenw: Please make it lower case as all logs are lower case.

mthenw:
Actually I will fix it in master. 

👍

tobyjoe:
No prob.

tj:
👍 thanks for the fix

PR #402, opened 2016-05-07
Parse Github tag JSON when the response has no linebreaks fixes #400

### LINKED ISSUE #400: Installer not working on Mac OSX
`install.sh` currently does not work on Mac. The JSON returned by `curl -s https://api.github.com/repos/apex/apex/tags` does not currently seem to include newlines, which causes [this command](https://github.com/apex/apex/blob/master/install.sh#L42) to fail to actually correctly parse the latest tag. The grep/sed commands need updated. The current output on my machine:

```
$ LATEST=$(curl -s https://api.github.com/repos/apex/apex/tags | grep name | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)
$ echo $LATEST
v0.8.0zipball_url
```

gmaclennan:
See #400 for discussion.


tj:
could we also just do a `-A ApexInstaller` to fix that?

gmaclennan:
With that user agent Github returns JSON with no linebreaks. Seems a little hit-and-mess whether Github includes linebreaks in their response.

tj:
k cool, maybe later we can use python or something ubiquitous, not sure if they have a stdlib json package or not

PR #403, opened 2016-05-08
add start of dynamic auto-completion

tj:
Add dynamic auto-completion so that commands accepting functions can properly complete based on the ./functions entries.

Still pending installation of the script, would love some input there. Options:

- add a command to do this
- add a prompt somewhere 
- try and automatically install when missing (preferrable)
- just leave the script in docs

```sh
_apex()  {
  COMPREPLY=()
  local cur="${COMP_WORDS[COMP_CWORD]}"
  local opts="$(apex autocomplete ${COMP_WORDS[@]:1})"
  COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
  return 0
}
complete -F _apex apex 
```

There might be a better way to get some of this from Cobra, seemed like it only generates a static script that you can use. Not sure if it's possible to delegate from that script for dynamic cases.


mthenw:
> try and automatically install when missing (preferrable)

Is it possible to make it reliable on *nix OSes?

tj:
Not sure, I wouldn't be surprised if they all have different directories

tj:
Ah I think we need a `--` in there for `apex autocomplete` haha

```
apex logs --fo   ⨯ Error: unknown flag: --fo
   ⨯ Error: unknown flag: --fo
   ⨯ Error: unknown flag: --fo
   ⨯ Error: unknown flag: --fo
   ⨯ Error: unknown flag: --fo
```

mthenw:
> Not sure, I wouldn't be surprised if they all have different directories

so maybe we should start with 

> just leave the script in docs

:)

tj:
haha 👍 works for now

mthenw:
LGTM!

tj:
- [ ] add `--`
- [ ] add docs for OSX / Ubuntu at least

Issue #288, opened 2016-03-01
Drop archive dep


tj:
Pretty unnecessary now, originally I think we were using AddDir.

PR #404, opened 2016-05-09
fix symlink support. Closes #401 #288

### LINKED ISSUE #401: Fix symlink support
With a simple `npm link` I get:

```
   ⨯ Error: function foo: Only regular files supported: node_modules/test
```

I thinkkkk the bulk of our logic in Function.Build for symlinks could be in utils.LoadFiles, since we'll need to read/walk directories, might as well resolve everything right there. Too tired to look into it tonight but maybe tomorrow

mthenw:
Important thing is that `AddDir` doesn't follow symlinks (the same reason as for `filepath.Walk`) so if there are nested symlinks it want work. Not sure it's OK(?)


tj:
sounds good to me, I haven't seen many people symlink within a node module much if ever

tobyjoe:
Having just touched this code last week, it looks good. Especially fond of the comment mentioning the symlink (potential) gotcha. If someone hits it in the future, it should be easy to find the spot to patch it.

tj:
LGTM as well, haven't tested yet but I'll do some QA stuff before release for a quick sanity check

tj:
awesome, works like a charm! <3

Node people can rejoice that `npm link` works haha

Issue #405, opened 2016-05-10
Disable versions cleanup


mthenw:
with `retainedVersions: -1`

PR #406, opened 2016-05-10
Fixes issue 405 - adds ability to disable version cleanup

tobyjoe:
This addresses the issue #405 - using a value of ```-1``` for ```RetainedVersions``` to disable the cleanup of deployed versions.

I'm opening a related issue around refactoring this behavior, but this one seemed clear. 

I also added some tests that allow us to test the private methods of the ```function``` package. If someone wants to do this in a better way, please modify or make suggestions! I just needed to get some coverage in for this logic since it's sort of... odd.

Finally, I added a default ```RetainedVersions``` value as a ```const``` in ```function``` and will separately propose that we do the same for a couple of ```const``` values in ```project```

🔬

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -613,10 +613,9 @@ func (f *Function) versions() ([]*lambda.FunctionConfiguration, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	versions := list.Versions
-	if *versions[0].Version == "$LATEST" {
-		versions = versions[1:] // remove $LATEST
-	}
+	versions = versions[1:] // remove $LATEST
 
 	return versions, nil
 }
--- a/function/function.go
+++ b/function/function.go
@@ -62,6 +62,9 @@ const (
 	RequestResponse InvocationType = "RequestResponse"
 	Event                          = "Event"
 	DryRun                         = "DryRun"
+
+	// DefaultRetainedVersions defines default number (i) of versions to retain on AWS
+	DefaultRetainedVersions = 10
 )
 
 // CurrentAlias name.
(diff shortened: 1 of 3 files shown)


function/function.go
```diff
@@ -600,8 +613,10 @@ func (f *Function) versions() ([]*lambda.FunctionConfiguration, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	versions := list.Versions[1:] // remove $LATEST
+	versions := list.Versions
+	if *versions[0].Version == "$LATEST" {
```
tj: I forget what the return value is for this versions call, wouldn't the latest _always_ be $LATEST?

tj:
LGTM other than the pointer thing you mentioned in the issue

function/function.go
tobyjoe: Actually, this is an area I was confused when trying to test ```Function``` in isolation of the whole ```Project.Open``` stuff.

tj: IIRC there's no need for the check

tobyjoe: I've removed that.

mthenw:
Is this PR still needed if there is https://github.com/apex/apex/pull/408?

tobyjoe:
@mthenw it depends on how you guys wanna go with it.

I submitted #405 to address that Issue, then submitted this one for discussion. If you guys are ok with some pointer vs value refactoring, I think there may be a few things I can shift in a different issue... 

But yeah, ultimately, you're right: this and #405 are mutually-exclusive.

Keep in mind that this one may be a "breaking change" because the value map is different (0 means zero, there is no -1, and default is nil and subs to 10). 

That would require at very least some prominent changelog visibility, so we may want to be conservative about it... Your call, guys!

PR #408, opened 2016-05-11
Fixes 407 retainedversions values

tobyjoe:
This switches the ```RetainedVersions``` bit to a pointer.

It's got passing tests, but I've had 🍻  and stand behind none of this. 😪


project/project.go
```diff
@@ -82,8 +82,9 @@ func (p *Project) defaults() {
 		p.NameTemplate = "{{.Project.Name}}_{{.Function.Name}}"
 	}
 
-	if p.RetainedVersions == 0 {
-		p.RetainedVersions = function.DefaultRetainedVersions
+	if p.RetainedVersions == nil {
+		tmp := function.DefaultRetainedVersions
```
tobyjoe: Is there a more idiomatic way to do this? Grabbing the address of a const?

function/function.go
```diff
@@ -600,8 +614,10 @@ func (f *Function) versions() ([]*lambda.FunctionConfiguration, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	versions := list.Versions[1:] // remove $LATEST
+	versions := list.Versions
+	if *versions[0].Version == "$LATEST" {
```
mthenw: is it really needed? $LATEST AFAIR is always on the top of that list.

project/project.go
```diff
@@ -82,8 +82,9 @@ func (p *Project) defaults() {
 		p.NameTemplate = "{{.Project.Name}}_{{.Function.Name}}"
 	}
 
-	if p.RetainedVersions == 0 {
-		p.RetainedVersions = 10
+	if p.RetainedVersions == nil {
+		tmp := function.DefaultRetainedVersions
+		p.RetainedVersions = &tmp
```
mthenw: it's better to use `aws.Int`

function/function.go
```diff
@@ -161,6 +164,11 @@ func (f *Function) defaults() {
 		f.VPC.SecurityGroups = []string{}
 	}
 
+	if f.RetainedVersions == nil {
+		tmp := DefaultRetainedVersions
```
mthenw: It's better to use `aws.Int`

function/function.go
tobyjoe: I think I need to look at these lines again. I based it on the #405 branch I made, and may have 🍻 'd it. Gimme a few hours to get to the office and I'll push a fix.

project/project.go
tobyjoe: never used that, but I'll switch it today! thanks for the pointer, @mthenw

mthenw:
LGTM, I will take another look tomorrow and merge.

function/function.go
```diff
@@ -59,8 +59,11 @@ type InvocationType string
 // Invocation types.
 const (
 	RequestResponse InvocationType = "RequestResponse"
-	Event                          = "Event"
-	DryRun                         = "DryRun"
+	Event = "Event"
+	DryRun = "DryRun"
+
+// DefaultRetainedVersions defines default number (i) of versions to retain on AWS
```
tj: nitpick: I think we can leave the comment out here since it looks weird with the others having nothing :D haha. I think the name DefaultRetainedVersions is nice and self-documenting anyway

function/function.go
```diff
@@ -447,8 +460,8 @@ func (f *Function) Rollback() error {
 		return errors.New("Can't rollback. Only one version deployed.")
 	}
 
-	latest := *versions[len(versions)-1].Version
-	prev := *versions[len(versions)-2].Version
+	latest := *versions[len(versions) - 1].Version
```
tj: hmm not gofmt-ing? It squishes operators like we have on the left

mthenw:
@tobyjoe LGTM, please fix TJ's comments.

tobyjoe:
Sorry, guys, was out at a conference for a week 🍹

PR #420, opened 2016-05-17
fix install script. Closes #419

### LINKED ISSUE #419: install instructions do nothing on Ubuntu 14.04
`curl https://raw.githubusercontent.com/apex/apex/master/install.sh | sh`

installation does not work on Ubuntu 14.04

mthenw:
On Ubuntu `$_` returns `/bin/sh`. I'm not bash expert so I've made it simpler :)


tj:
hrrm seems fine to me I use it all the time:

```
ubuntu@web-01:~$ curl -sL https://github.com/apex/apex/releases/download/$LATEST/apex_$PLATFORM -o /usr/local/bin/apex
ubuntu@web-01:~$ echo $_
/usr/local/bin/apex
```

tj:
doesn't hurt to be less obscure though haha, bash is a hot mess. mergin!

tj:
interesting, guess I misinterpreted that var :D. good ol bash

0xmohit:
In this case, `_` expands to the last argument to the previous command, after expansion.

More in the [manual](https://www.gnu.org/software/bash/manual/bash.html#Special-Parameter)

tj:
The way it's describe still sounds like `/usr/local/bin/apex` to me, seems odd, but yea this is certainly more clear haha

PR #424, opened 2016-05-22
Allow usage of method names other than handle for Python

abhiyerra:


tj:
hmm this should be taken care of by: https://github.com/apex/apex/blob/master/plugins/python/python.go#L30 and  https://github.com/apex/apex/blob/master/plugins/python/python.go#L77

I'm not too familiar with python, I think @mthenw wrote the python env stuff. The basic idea here is that when environment variables are used we always have a constant handler (_apex_main.handle) that just invokes the other, whatever its name was

abhiyerra:
I guess that doesn't happen if I have a `handler` in say `index.py` with the full handler being `index.hander`. _apex_main.handle will include `from index import handler`. Maybe I didn't realize that the behavior is to always name the handler method `handle`. If that is the case if might be better to include it in the documentation since I was pulling my hair a little bit trying to figure out what was happening.

mthenw:
For sure there is something wrong here. I will take a look today.

mthenw:
You are right. In python user's func is imported tho the main/wrapper file. Thanks for that!

tj:
ah I see, makes sense! I was thinking of our node case where it's just exported as 'handle': https://github.com/apex/apex/blob/master/plugins/nodejs/nodejs.go#L36

I guess python doesn't have that haha

PR #431, opened 2016-05-30
fix panic when 'apex invoke -L' throws an error

jonnangle:
`fn.Invoke` returns `nil` for logs when an error occurs


tj:
thanks!

PR #432, opened 2016-05-31
add multiple deployment targets for different environments

filiphaftek:


kszarek:
I'm looking forward to see it in master branch 👍

kszarlej:
would be great to have this feature in mainline apex :)

tj:
I'd like to use `--env` for this, but I think rather than having a bunch of dirs and files I'd like something like:

```json
{
  "name": "test",
  "runtime": "nodejs",
  "memory": 128,
  "timeout": 5,
  "envs": {
    "production"" {
      "memory": 1024 
    }
  }
}
```

same with project.json. Though for functions I'd argue they should always match production, but you could also have prod-specific env vars in there I suppose

annamikulinska:
@tj so what do you think - it would be ok to have function.development.json and function.production.json than ?

tzaleski:
seems like a nice feature 👍

DavidTPate:
I think having separate files would be preferred. The reason why is that with them being in the same file I could eventually see people wanting things to override things by file. Whereas when the files are separate they are clearly different.

Additionally, it would provide us with the ability to change all the properties of the `project.json` for each deployment environment, while also not causing the additional nesting that @tj alluded to. My defaults for Security Groups for example vary by each environment.

filiphaftek:
I agree with you @tj that webpack approach with multiple configuration files is better than nested directories and that --env is exactly the flag which should be used so I modified the PR to enable multiple project and function files usage. I decided not to fallback to project.json or function.json file in case when you add --env ENV and file project.ENV.json or function.ENV.json does not exist, as I assumed that if apex user wants to use multiple environments he should create these separate files.

tj:
Hmm yea that's tricky, need to give it some more thought. I'd still like to make things simple for people who are just getting started, then when they need multiple envs they can add that. We still need an entry-point for default anyway, so we can't assume ./project.stage.json etc

Right now the main problems I think are:

- users should be able to specify a default env (usually stage) so you don't need a ton of flags just to get started. Right now we have `defaultEnvironment` in project.json for `apex infra`

- `--env NAME` and `--profile NAME` are kind of similar ATM, but if you have to do `apex deploy --profile myapp-prod --env prod` then that's just bad haha. I wouldn't mind allowing you to specify profile names in the JSON config so only `--env prod` is necessary

tj:
I think regarding config I'd personally just use single project/function.json files, nesting might make a bit more sense since it enforces a default environment. If people prefer breaking that out into multiple maybe we could support both

filiphaftek:
@tj answering to problems you added in comment:
- 'don't need a ton of flags' - with proposed approach, you can always use default env and project.json/function.json for starting user, as my code shouldn't break the backward compatibility (maybe my project structure in Readme wasn't enough clear:)
- '--env and --profile are similar' - I partially agree, as profile is just the user credentials for login and env is the application working environment. It would be easier for apex user to have single --env for application and add profile to json file. For me first usage of --profile was a bit misleading :)
- 'nesting might make a bit more sense since it enforces a default environment' - IMHO nesting is more complicated and makes merging harder, so I would vote for separate files ;)

XanderDwyl:
looking forward to see this on production... 👍

tj:
sorry for the delays I'll try and dig into this soon. Just need to make sure everything compliments existing features etc, easy to add things, hard to take them away! Hopefully this week I'll have time to review/tweak/document.

 I'm still not entirely sold on separate files personally, at least not exclusively, since the default behaviour is what we currently have documented it feels weird go from all function.json / project.json files to function.<env>.js and project.<env>.js files. I'll think some more about the flags too so it's less redundant

binarydud:
it might be nice to note in the docs that this breaks older projects that might have been using the -e flag in tandem with the infrastructure/ directory.

tj:
@binarydud ah you're right, sorry about that. That's buggy, I think we should still load project.json if an env specific one does not exist, I'll open an issue for that

PR #442, opened 2016-06-08
add env configuration note to docs. Closes #436

### LINKED ISSUE #436: Error: json: cannot unmarshal number into Go value of type string
In my `project.json` file I have the config like the following: 
```javascript
{
    name: xxx,
    description: xxx,
    memory: 128,
    timeout: 10,
    role: arn:aws:iam::xxx:role/xxx,
    defaultEnvironment: dev,
    environment: {
      AWS_REGION: "us-west-2",
      DEFAULT_TOKEN_SIZE: 12,
    }
  }
```
and when I run `apex deploy`, I get the error says `json: cannot unmarshal number into Go value of type string`, it turns out that I have to make a string value for `DEFAULT_TOKEN_SIZE` instead of a integer.

I am wondering it apex should support the integer value for environment config inside `project.json`

mooyoul:
🍻

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -26,3 +26,5 @@ Specify environment variables in project.json or function.json:
   }
 }
 ```
+
+Environment variables in json configuration are must be string.


tj:
:D

PR #457, opened 2016-06-16
add aliases to apex list. Closes #430

### LINKED ISSUE #430: Display aliases on `apex list`
Related to #364, if you use aliases as a sort of tagging system even just for versioning purposes, it would be nice to list them in `apex list` so you can see what has been released etc.

jonnangle:
```
$ apex list

  test
    runtime: nodejs
    memory: 128mb
    timeout: 5s
    role: arn:aws:iam::123412341234:role/MyRole
    handler: index.handle
    aliases: current@v6, foo@v5
    current version: 6
```

I wasn't too sure how to present the aliases; this seems OK, but showing the `current` alias kinda duplicates the information from `current version:` on the next line. let me know what you think :)

### DIFF
--- a/cmd/apex/list/list.go
+++ b/cmd/apex/list/list.go
@@ -78,7 +78,7 @@ func outputTFvars() {
 func outputList() {
 	fmt.Println()
 	for _, fn := range root.Project.Functions {
-		config, err := fn.GetConfigCurrent()
+		_, err := fn.GetConfigCurrent()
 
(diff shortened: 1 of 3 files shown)


tj:
nice! yeah I agree, we could probably remove `current version`

jonnangle:
I've done that, and updated the doc.

tj:
great thanks!

PR #475, opened 2016-07-07
add aws-sdk workaround to webpack example

tlvince:
Using aws-sdk in the context of Node-based Lambda functions is common, so
including this will save a headache for the next person.


tj:
ahh interesting thanks! That does not sound fun

Issue #491, opened 2016-07-14
Hooks not working for java runtime


akranga:
When building java runtime function, then hooks not triggered. 

Steps to reproduce:
1. Create `function.json` file: 
```json
{
	"runtime" : "java",
	"hooks": {
		"build": "./gradlew jar",
		"clean": "./gradlew clean"
	}
}
```

2. Run `apex deploy <fn_name>`
3. See error: `Error: function ui: Expected jar file not found`
Expected Result:
Gradle build should be triggered prior function deployment

tj:
Hmm the java plugin was contributed, looks like you're hitting https://github.com/apex/apex/blob/master/plugins/java/java.go#L70. I don't know anything about java idioms unfortunately so I can't comment on what might be wrong there, `-l debug` might give you some more insight for what it's trying to find

akranga:
Yeah, I was running with -l debug. Pushed PR with pretty trivial fix

tj:
awesome thanks for the fix!

noerremark:
I had a similar issue. Turns out that maven wasn't installed...

PR #492, opened 2016-07-14
run hooks before runtime plugins to fix (#491)

akranga:
Fix is pretty trivial. To rearrange array of plugins. Then gradle build has been hooked before deploying function


akranga:
Build failed on style check. It seems unrelated to change

tj:
ahhh nice catch!

PR #495, opened 2016-07-16
fix #195, handles JSON array as input to invoke

### LINKED ISSUE #195: Handle array in request.json
Suppose,
I have following the request.json file

```javascript
[
   {
      "event":"processed"
   },
   {
      "event":"deferred",
   }
]
```

and then I invoke it with lambda function:
```sh
$ cat request.json | apex invoke --logs webhook
⨯ error parsing response: json: cannot unmarshal array into Go value of type map[string]interface {}
```

I immediately get the error from apex invoke. Wondering how do we support invoking of JSON arrays?

pkazmierczak:
my very first golang pull request, should probably be reviewed
carefully…


tj:
haha :D. The use-case is `[event, event]`? Personally I think I would prefer to keep the regular single event input, because you can still stream if you have `events | apex invoke foo`, as long as `events` outputs say:

```
{ event one }
{ event two }
```

pkazmierczak:
@tj sure, I just read the discussion in #195 and thought you'd like both solutions to work.

tj:
ahhh hmm I should close that one, I thought maybe they wanted that entire array to end up in the function but reading it now, I think they just wanted to do multiple invokes

PR #504, opened 2016-07-26
expose function-specific role to infra

ajcrites:
Function-specific roles are currently not passed to terraform -- only the project role is passed (if it exists).

Now, `apex_function_$FNNAME_role` will be passed with the corresponding function role. This is also the project role, if it is available.

### DIFF
--- a/infra/infra.go
+++ b/infra/infra.go
@@ -67,6 +67,11 @@ func (p *Proxy) functionVars() (args []string) {
 
 		args = append(args, "-var")
 		args = append(args, fmt.Sprintf("apex_function_%s=%s", fn.Name, *config.Configuration.FunctionArn))
+
+		if fn.Role != "" {
+			args = append(args, "-var")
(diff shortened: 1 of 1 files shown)


tj:
I forget why we even have apex_function_role TBH haha. Once we have the interpolation support (with func for grabbing ARNs etc from Terraform) it should get rid of the need for this right?

ajcrites:
Can you explain more about the interpolation support?

Right now I have apex called through a terraform provision (terraform creates the role) and the provision calls apex infra to set the triggers. If I needed to use the role ARNs in the infra this would be useful. I don't think I do right now but it could be handy for doing something like attaching a new policy to the role.

tj:
It's basically the same stuff TF provides, one thing I want it for is to do something like `"description": "my function ${shell('git describe --always')}"`, but we could have a func for grabbing TF state (so you can put the roles in TF), then there's no need to grab the roles from Apex, they would already be in TF.

DavidTPate:
Yeah, it seems to me that the interpolation support would solve this @tj

PR #514, opened 2016-08-04
Allow use of IAM roles

DaveBlooman:
From https://twitter.com/dblooman/status/760878975392579584.  

Use case is having 2 AWS accounts, dev/prod, with the desire to deploy them in a CI environment without issuing AWS keys directly.  This will allow for a `--iamrole` CLI flag to contain a role arn such as `arn:aws:iam::000000000:role/lambdassume`.


DavidTPate:
Specifying the role is already support within the `project.json` file.  Take a look here: http://apex.run/#multiple-environments

Apex also supports grabbing the role from Terraform automatically for each environment.

DaveBlooman:
@DavidTPate It is my understanding that the role in project.json is a global role for all lambda functions, not a role to assume when creating resources within AWS.

DavidTPate:
Ah, I understood your issue incorrectly @DaveBlooman. So let me try again. The issue you are having is not with the `role` that Lambda assumes but instead the role that is used to setup the function and retrieve logs and everything. 

So currently Apex supports using `profiles` specified for AWS, as detailed [here](http://apex.run/#via-aws-files). These profiles could be Roles, Credentials, or other forms of authentication that AWS supports. 

Here's an [example](http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html) of a profile that Assumes a role, or other ones [here](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-multiple-profiles).

You can also specify the `profile` flag in your `project.json` or `project.prod.json` which points to the profile name for each environment.

DaveBlooman:
One of the things I was trying not to do was to hard code the roles on the EC2 instance.  As it stands, I would need to rebuild and deploy my build agent to add a new role, which takes effort and time in order to not impact daily dev workflow.  I would argue that specifying role ARNs is a more common way to handle CI environments, which most plugins offer.

tj:
Weird I swear the SDK has AssumeRoleProvider in the default chain picking it up from the EC2 metadata, I think your CI use-case makes sense. It's not something I use personally so it might be a burden to QA, but if it's already in the SDK or simple to add there I'm happy to have it. I think we'd need to handle the expiration as well (unless it does that internally).

DaveBlooman:
The SDK does pick up roles that are available on boot, but not roles available to the instance to be assumed.  Most CI plugins that tie into AWS work in this manner, here is an example on how the CodeDeploy plugin for Team City works

screenshot 2016-08-08 22 24 32

The default duration is 15 minutes, but I have set it at 30.  The only scenario I can see where this would be an issue is if you had a lot of functions with very long builds/zips.  In that scenario, the amount of time for Apex to package an upload your code would depend on network, capabilities of server etc.  

As far as QA goes, I can take on responsibility for ensuring it works, it already is a part of our build pipeline for deploying production functions.

tj:
Ahh I see I was thinking of the EC2 instance role, different case I guess! Wonder if we could add this to the cred chain as well and have `AWS_ROLE` or something to keep it more inline with the others. Seems like something each framework shouldn't really have to worry about

DaveBlooman:
AWS_ROLE env could work quite nicely, in most CI's you can pass env vars to the build job.  I guess if we follow the same pattern already in apex, people can over ride the env with the flag.  I'll add that in and test it

DaveBlooman:
If you look at most of the CI plugins out there, there is basically the same code to implement assume role and get creds , which is actually why a lot of people use access keys in plugins instead.

I've updated and tested on my end, works great with `AWS_ROLE` env var

tj:
awesome! LGTM, if you don't mind a little mention in https://github.com/apex/apex/blob/master/docs/aws-credentials.md for the env/flag & CI case would be rad then we should be good to merge

DaveBlooman:
Let me know if that is enough on the docs side, clear enough etc

DaveBlooman:
Ready to merge :)

DavidTPate:
@DaveBlooman Looks like the CI builds are failing due to Lint issues.

DaveBlooman:
@DavidTPate Hmm, not sure what to do here, nothing I've touched.

DavidTPate:
@DaveBlooman Yeah, sorry about that I noticed the same thing in a documentation change that someone submitted. Thanks for checking!

DaveBlooman:
@DavidTPate Good to merge then?

DaveBlooman:
@DavidTPate Good to merge then?

skout23:
So seems like **all** builds of apex on semaphoreci are failing cause of the reported output from the CI's call to gometalinter.  
`gometalinter --skip=mock --skip=boilerplate --disable-all --enable=golint ./...` 

These fails are caused by a change in gometalinter as outlined in this issue; https://github.com/alecthomas/gometalinter/issues/147  

The reports are all from the mock/ directory, which gometalinter is supposed to skip based on the passed flags.

A quick fix would be to just call golint directly and ignore the output from "skipped" directories.  `golint ./... | grep -v "[^mock\|^boilerplate]"` 

I don't know if you have the option to change this step in the CI, but if so this would get the apex builds passing again, and ideally would get this PR merged.

Issue #526, opened 2016-08-16
Add Swift runtime


paulofaria:
Now that Swift runs on the server having a Swift runtime would be an amazing addition for Apex.

balupton:
dupe of https://github.com/apex/apex/issues/86

PR #527, opened 2016-08-16
[WIP] Add Swift runtime

paulofaria:
This PR addreses #526 adding Swift runtime. Currently Swift doesn't support cross compiling. So unfortunately the plugin only works on Linux.


_examples/swift/functions/uppercase/.apexignore
```diff
@@ -0,0 +1,3 @@
+*.swift
+.build
+Packages
```
paulofaria: I'm no sure these work. Ignoring directories. Unfortunately I had to remove them manually in the build hook, which is quite ugly.

plugins/swift/swift.go
```diff
@@ -0,0 +1,39 @@
...
+type Plugin struct{}
+
+// Open adds the shim and golang defaults.
+func (p *Plugin) Open(fn *function.Function) error {
+	if fn.Runtime != Runtime {
+		return nil
+	}
+
+	if fn.Hooks.Build == "" {
+		fn.Hooks.Build = "swift build -c release; mv .build/release/main main; mv .build/release/*.so .; rm -rf .build; rm -rf Packages"
```
paulofaria: This is quite ugly. But I'm not sure if there's a better way.

_examples/swift/project.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "swift",
+  "runtime": "swift",
```
paulofaria: Is there a way to autodetect the Swift environment? I couldn't find a way to do it.

tayhalla:
the resulting binary will still try to call shared swift libs not present on the aws linux ami... have you tested this? Also, its not just linux that's the issue, it's that the swift compiler released under swift is for Ubuntu. The lambda AMIs are the AWS ECS ones, which has a AWS proprietary form of linux derived from REHL/Cent.

paulofaria:
I tested the example project and it worked fine. The Swift libraries have to be on the function directory to get them included in the zip. The plugin could get them from a known location and put them in the zip. This way the function directory only needs `main.swift` and `Package.swift`. The problem is that on macOS there is a known directory for the Swift libraries. on Linux we don't have that. So the plugin could check for a known location for the plugin. If there's nothing there it downloads the libraries and put them there.

paulofaria:
Cool! Now the plugin downloads the Swift libraries to `.apex/plugins/swift` if they haven't been downloaded yet. On the next builds it gets the libs from there. Also added inference for Swift functions. I'm not a go programmer so I don't know if the code is idiomatic enough. Would love some input on that. @tj

paulofaria:
I'll have some time this weekend to finish this off. Just so you know this PR isn't dead, haha.

lexrus:
Since Docker now has a native environment in macOS, how about compiling/running swift files in a docker container?
https://github.com/swiftdocker/docker-swift
I've tested compile a helloworld.swift with this docker image and copy the necessary *.so files and successfully run it in AWS Lambda.

paulofaria:
That's a good idea! how do you feel about updating this PR to add that feature?

tj:
Docker stuff might be ok, as long as we make it super obvious. That would make C/C++/Rust much easier as well, sucks to have the dep but hey!

paulofaria:
@lexrus ping!

lexrus:
@paulofaria I'm satisfied writing Lambda in Go lang these days. 😄

choefele:
Hi – I was wondering if I can help adding Swift support to Apex. I've successfully published an Alexa Skill implemented in Swift that runs on Lambda using the following approach:
1. Build Swift executable in Docker container based on Ubuntu (first Docker command in [`run-integration-tests.sh`](https://github.com/choefele/swift-lambda-app/blob/master/run-integration-tests.sh))
2. Extract and copy the app's library dependencies via `ldd` on Ubuntu (second Docker command in [`run-integration-tests.sh`](https://github.com/choefele/swift-lambda-app/blob/master/run-integration-tests.sh))
3. Run Swift executable on Amazon Linux via `ld-linux` from a Node.js shim – this makes sure that the glibc version matches exactly the one needed ([index.js](https://github.com/choefele/swift-lambda-app/blob/master/Shim/index.js))

There's a [sample app](https://github.com/choefele/swift-lambda-app) and a more detailed description of my approach explained in [Serverless Swift](https://medium.com/@claushoefele/serverless-swift-2e8dce589b68#.izas0ic35).

If I understand correctly, adding a Docker build step would be ok with the project maintainers. If so: is there more documentation on how to build the Apex command locally and interface the Swift app needs to implement to interface with the existing shim?

Thanks

paulofaria:
@choefele if you take a look at this PR you can understand how plugins work. it's quite easy actually. if you need anything let me know.

Issue #524, opened 2016-08-16
unable to init a lambda using a profile that contains a role_arn


luck02:
- __Operating system__:  OSX 10.11.6
  - __Apex Version__: 0.10.2

As far as I can tell my credentials and config file are correct.  when I execute:

```shell
⇒  aws iam create-role --profile data --role-name test-this-role-delete-when-done --assume-role-policy-document file://policy-doc.json
```

I get the expected output (role is created).  note I'm using the --profile data flag.

So I'm going to assume my AWS config is demonstrably correct

When I execute:

```shell
⇒  apex --profile data init -ldebug


             _    ____  _______  __
            / \  |  _ \| ____\ \/ /
           / _ \ | |_) |  _|  \  /
          / ___ \|  __/| |___ /  \
         /_/   \_\_|   |_____/_/\_\



  Enter the name of your project. It should be machine-friendly, as this
  is used to prefix your functions in Lambda.

    Project name: proj_name

  Enter an optional description of your project.

    Project description:

  [+] creating IAM ub-data-event-bus-persistor_lambda_function role
   ⨯ Error: creating role: NoCredentialProviders: no valid providers in chain. Deprecated.
       	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

When I execute this command without the --profile data flag I get:

```shell
  [+] creating IAM ub-data-event-bus-persistor_lambda_function role
   ⨯ Error: creating role: AccessDenied: User: arn:aws:iam::99999999:user/gary.lucas is not authorized to perform: iam:CreateRole on resource: arn:aws:iam::99999999:role/lambda_user
       	status code: 403, request id: 93617ee6-63a6-11e6-a945-c1fc127febbb
```

My default role does not have those required permissions so that's expected.

My credentials file is standard and set up via aws configure command.  My config file looks more or less like this:
```shell
[default]
region = us-east-1
output = json

[profile data]
role_arn = arn:aws:iam::999999:role/a-valid-cross-accountRole
source_profile = default
output = json
region = us-east-1
```

Note that I had to add region to profile data as apex wasn't picking it up.

My assumption (untested) is that apex is just issueing commands via the AWS SDK so it should work.  But perhaps that's not the case?  Does anyone have an experience using roles that worked?

DavidTPate:
[...]

PR #529, opened 2016-08-17
Add a minimum IAM policy to the documentation

DavidTPate:
This PR adds a minimum IAM policy needed for Apex to run. This was generated by retrieving the calls to the AWS SDK and verified by testing a role with the permissions to ensure it was the minimum necessary being listed here.

Please suggest if there are some that I missed, but I think I covered most cases with my testing.

This helps to fix one request that was part of #524


rowanu:
Can we get this merged please? Would be nice to see this info in the docs.

DavidTPate:
You know I completely forgot about this older PR of mine. Thanks @rowanu I'll get it in there.

DavidTPate:
Merging this, as it only a document change and the build is failing for other reasons (linting).

PR #530, opened 2016-08-17
fix config comparison function. Closes #528

### LINKED ISSUE #528: Apex deploy not idempotent in either config or code
Apex version 0.10.2
Apex platform: Go
Mac OSX Yosemite 10.10.5

When I try to `apex deploy` new versions of the code are _always_ deployed, the config is detected as if it has changed as well.  Even in back-to-back calls to `apex deploy` functions and configs are updated.  I also deployed 2 different versions, downloaded the zip files from the lambda UI and diffed the files - they did not differ.

** Info I am not sure whether it is relevant or not **
* I didn't notice this when I did the apex hello world example
* I do not have a function.json file
* ".go" files are apexignored
* I am running in a VPC
* I have 2 "instances" of apex running on my AWS account in order to isolate different IAM roles necessary
* I have added custom defined policies to my lambda role
* I have all of my lambda functions connected to CloudWatch events

**Project.json**
```json
{
  "name": "infra",
  "description": "This is a set of functions that will be used to manage infra",
  "runtime": "golang",
  "memory": 128,
  "timeout": 120,
  "role": "arn:aws:iam::AWS_ACCOUNT_ID:role/infra_lambda_function",
  "environment": {},
  "vpc": {
    "securityGroups": [
      "sg-SG1"
    ],
    "subnets": [
      "subnet-A",
      "subnet-B",
      "subnet-C"
    ]
  }
}
```

**.apexignore**
```
.*go
```

`apex deploy trigger_buildbot -l debug`
```text
   • loading functions in functions
   • loading function in functions/trigger_buildbot
   • deploying 1 functions    
   • deploying                 function=trigger_buildbot
   • creating build            function=trigger_buildbot
   • injecting prelude         function=trigger_buildbot
   • hook                      command=GOOS=linux GOARCH=amd64 go build -o main *.go function=trigger_buildbot hook=build
   • adding env                env=map[LAMBDA_FUNCTION_NAME:infra_trigger_buildbot APEX_FUNCTION_NAME:trigger_buildbot] function=trigger_buildbot
   • add shim                  function=trigger_buildbot
   • add file to zip           file=main function=trigger_buildbot
   • created build (2.6 MB)    function=trigger_buildbot
   • fetching config           function=trigger_buildbot
   • config changed            function=trigger_buildbot
   • updating config           function=trigger_buildbot
   • updating function         function=trigger_buildbot
   • updated alias current     function=trigger_buildbot version=20
   • function updated          function=trigger_buildbot name=infra_trigger_buildbot version=20
   • cleaning up version: 10   function=trigger_buildbot
   • cleaning 1 functions     
   • hook                      command=rm -f main function=trigger_buildbot hook=clean
   • conditional flush         above_duration=24h0m0s above_size=500 age=38m33.282169122s size=73
   • close     
```

I also notice this warning in Lambda:

screenshot 2016-08-17 00 31 16

skabbes:
When comparing the subnets and security groups for remote and local
configs, be sure to treat them as sets - since that's how AWS treats
them.


function/function.go
```diff
@@ -753,6 +754,12 @@ func (f *Function) configChanged(config *lambda.GetFunctionOutput) bool {
 		}
 	}
 
+	// don't make any assumptions about the order AWS stores the subnets or security groups
+	sort.StringSlice(remoteConfig.VPC.Subnets).Sort()
```
DavidTPate: What about using an actual `Set` instead? It looks like we could use something like [golang-set](https://github.com/deckarep/golang-set) to add the ability to construct Sets in Go and allow that to handle this behind the scenes.

tj:
LGTM

function/function.go
tj: Better to stick with the stdlib I think, this looks fine to me

DavidTPate: Sounds good!

skabbes:
Thanks for the quick review, and awesome tool.

function/function.go
skabbes: I'd echo tj on that too.  It is the "go way" :).

Issue #481, opened 2016-07-12
.apexignore does not match folder


andreineculau:
If you have a bug report please let us know the following:

  - __Operating system__: OSX 10.11
  - __Apex Version__: 0.10.2

This is an issue with https://github.com/rliebling/gitignorer probably, but here goes: .apexignore does not ignore dirs specified without a trailing slash e.g. `folder_to_ignore` in .apexignore would still pack `folder_to_ignore/some.file`.

These work though: `folder_to_ignore/`, `folder_to_ignore/*`, `folder_to_ignore/**`.

I'm creating .apexignore programmatically, thus I could do without an extra file/folder check.

tj:
I'll have a look! Marking as a bug

PR #550, opened 2016-09-13
Fix .apexignore fails when folder without trailling slash

doomsplayer:
I think this fixes issue #481


DavidTPate:
@doomsplayer Could you add some tests to the test suite to confirm that this works as expected and that we don't regress please?

doomsplayer:
finished
@DavidTPate

PR #564, opened 2016-09-30
Building fails java-example - hook and runtime missing

noerremark:
Just a minor fix that makes sure building actually works.

### DIFF
--- a/_examples/java/functions/with-pom/function.json
+++ b/_examples/java/functions/with-pom/function.json
@@ -1,4 +1,9 @@
 {
   "description": "Java example function with existing pom.xml",
-  "handler": "lambda.Example::handler"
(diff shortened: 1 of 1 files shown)


tj:
hmm something must have changed, we used to infer the runtime https://github.com/apex/apex/blob/master/plugins/inference/inference.go#L21, but it wouldn't hurt to be explicit anyway. Builds ideally work out of the box as well, but I'm not at all familiar with Java idioms so maybe there's no single answer there

noerremark:
@tj It didn't work for me, and I like this project, so I thought that I would contrib the fix.

tj:
Ideally with fix that stuff, but this works for now! Thanks

PR #567, opened 2016-10-06
added apex environment awarness to hook command (fix #566)

### LINKED ISSUE #566: pass apex env variables to hook command
As of now apex support hooks. However hooked build process often needs context information about apex execution. 

Proposition to propagate apex env to hook command as environment variables.

akranga:
Now apex environment and function environment variables has been propagated to hook command as env_vars. So, downstream build (triggered by hook) can be aware of upstream apex


DavidTPate:
@akranga Could you add some tests to verify this functionality please?

plugins/hooks/hooks.go
```diff
@@ -57,8 +57,20 @@ func (p *Plugin) run(hook, command string, fn *function.Function) error {
 		"command": command,
 	}).Debug("hook")
 
+	convertToEnviron := func(env map[string]string) ([] string)  {
```
tj: I'd move this to utils since we don't need a closure

function/function.go
```diff
@@ -112,6 +112,7 @@ type Function struct {
 // Open the function.json file and prime the config.
 func (f *Function) Open(environment string) error {
 	f.defaults()
+	f.Setenv("APEX_ENVIRONMENT", environment)
```
tj: maybe just APEX_ENV, unless we have this somewhere else, can't remember :D

akranga: I used the same name as for terraform

tj: ahh ok cool, I thought we might already have that name

matthewmueller:
Does this also support environment switching via an environment variable, ie.  `APEX_ENVIRONMENT=prod apex deploy`?

PR #568, opened 2016-10-07
Updated python dependency example to use build hooks

sagarsahil:
Updated the example to install the dependencies using the build hook. Existing example was not clear and i saw example of lot of users manually copying the package.


tj:
Awesome thanks!

PR #574, opened 2016-10-22
Document that Function ARNs are Not Available to Terraform Until After a Deploy

chrisguitarguy:
This detail 100% makes sense, but can be a bit confusing for someone first using Apex + Terraform to provision things for the first time.

I also updated the [terraform interpolation syntax](https://www.terraform.io/docs/configuration/interpolation.html) to `${var.thing}` rather than `$(var.thing)`.


tj:
thanks!

PR #591, opened 2016-11-10
fix babel-webpack example, added nodejs4.3 and webpack2 example

thewillhuang:
Was playing around with babel-webpack example and found out that it was missing babel-core which is required for the example to work.

Also made a webpack2 / nodejs4.3 example branch with an updated readme for that branch


tj:
LGTM thanks!

PR #592, opened 2016-11-10
allow override of $DEST

joekiller:
If $DEST is set, install apex to that destination.


tj:
hmm maybe we should do PREFIX= which is more conventional, but LGTM for now!

joekiller:
Thank you!

PR #598, opened 2016-11-18
fix overrides of VPC in function.json.

### LINKED ISSUE #597: VPC config in a function clobbers VPC defaults in a project
I have a project with several functions. In the `project.json` I define the `securityGroups` I want to use as the defaults for functions. However, if I override the `securityGroups` in any function, then when I do a deploy of the entire project, any function that get deployed after the function with the override will get the `securityGroups` of the overridden function.

Example: 

```
# project.json
{
  ...
  "vpc": {
    "subnets": [
      "subnet-xxx",
      "subnet-yyy"
    ],
    "securityGroups": [
      "sg-xxx",
      "sg-yyy"
    ]
  },
  "runtime": "python2.7"
}
```

```
# functions/a/function.json
{
  "handler": "a.main",
  "vpc": {
      "subnets": [
          "subnet-xxx",
          "subnet-yyy"
      ],
      "securityGroups": [
          "sg-xxx",
          "sg-zzz"
      ]
  }
}
```

```
# functions/b/function.json
{
  "handler": "b.main"
}
```

Now if I run `apex deploy b`, then `b` ends up with the expected default security groups of `sg-xxx` and `sg-yyy`. However if I run `apex deploy`, function `b` ends up with the security groups defined in function `a`.

I'm not very familiar with go, but it seems to me like the `VPC` object from the project config is [passed into the function config](https://github.com/apex/apex/blob/master/project/project.go#L350), and the function then [overwrites the members of the `VPC` object](https://github.com/apex/apex/blob/master/function/function.go#L163), rather than replacing it with a new `VPC` object.

zroger:
The project's VPC needs to be deep copied before passing it to the
function, otherwise overrides modify the project's settings.

Fixes #597.

### DIFF
--- a/project/_fixtures/vpcOverride/functions/bar/function.json
+++ b/project/_fixtures/vpcOverride/functions/bar/function.json
@@ -1,0 +1,7 @@
+{
+  "vpc": {
+    "securityGroups": [
+      "sg-override"
+    ]
+  }
+}
--- a/project/_fixtures/vpcOverride/project.json
+++ b/project/_fixtures/vpcOverride/project.json
@@ -1,0 +1,9 @@
+{
+  "name": "vpcOverride",
+  "role": "testrole",
+  "vpc": {
+    "securityGroups": [
+      "sg-default"
+    ]
+  }
+}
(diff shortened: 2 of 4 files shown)


tj:
awesome, wonder if reflect has copying, I imagine we'll have this with some of the other props as well

PR #614, opened 2016-12-12
Allow use of IAM roles Part 2

DaveBlooman:
I deleted my old fork a while ago, so PR #514 won't update, but there is an update PR with the latest master.


cmd/apex/root/root.go
```diff
@@ -130,6 +136,32 @@ func Prepare(c *cobra.Command, args []string) error {
...
+		if err != nil {
+			return err
+		}
+
+		Config = &aws.Config{
+			Region:      aws.String(region),
+			Credentials: credentials.NewStaticCredentials(*stsresp.Credentials.AccessKeyId, *stsresp.Credentials.SecretAccessKey, *stsresp.Credentials.SessionToken),
+		}
+	}
+
```
tj: Any thoughts on us maybe moving some of this into ./utils? Most Go people hate util packages haha but I'd rather stuff things in there and keep inline to a minimum

cmd/apex/root/root.go
```diff
@@ -130,6 +136,32 @@ func Prepare(c *cobra.Command, args []string) error {
 		environment = os.Getenv("APEX_ENVIRONMENT")
 	}
 
+	// iamrole from flag, env
+	if iamrole == "" {
+		iamrole = os.Getenv("AWS_ROLE")
+	}
+
+	if iamrole != "" {
+		stscreds := sts.New(session.New(&aws.Config{Region: aws.String(region)}))
```
tj: I think we'd be safe to re-use the `Config` above

cmd/apex/root/root.go
DaveBlooman: I guess in this case, we add a dependency onto the utils package, but, I'm not that bothered.

tj: The key thing for me at least, is to have shorter functions so it's clear where the boundaries are, plus it keeps the commands a bit more declarative which is nice

DaveBlooman: Sounds good, I made a slight tweak to completely build the config object in the utils package, should be good to go.

PR #622, opened 2016-12-19
Add exec command Closes #619

### LINKED ISSUE #619: Discussion: Is the Terraform usage model for Apex 'right'?
I've been trying to get a solid Terraform workflow running with Apex, but I'm hitting a few stumbling issues. It seems to me the prevalent way to use this is conjunction with something like API Gateway, but it feels like structurally the setup might be wrong. In short, because the terraform scripts live in-situ with the code it means that you need to either take on the overhead of TF modules or you have to repeatedly duplicate your Terraform files: this means stuff like state for API gateway definitions is a nightmare, and you've potentially got 3x set of scripts to keep in line - plus then how do you cleanly promote an API gateway definition across it's stages?

I think that having the Terraform scripts themselves be physically different is problematic: the per environment customisation can already be done pretty well with tokens and variables in a single set of scripts generally. There's also problems where the role ARN we use needs to be provisioned per-environment too, and we cant just use the as-is implementation. I basically run apex infra twice, once with a filter for specific assets, then again after an apex deploy.

_Proposed_: Drop Terraform specific support (deprecate, show a message on how to migrate with a link, then drop in 2-3 versions time), and have an environment variable passthrough function - so when you do:

- apex run -e env -dir workdir terraform tf-params-here

What happens is that it shells through, but setting various APEX_FUNCTION_name=ARN, APEX_ENV=blah variables as today - but *completely decoupled from Terraform* and its var models, so it can be used with other tools transparently. It means a shift from ${var} to ${env} - but then makes the whole process tool agnostic.

Then people like me can do effectively custom setups without needing Apex to support every permutation. I can just have a shell script that runs 3x Apex commands, so I can..

- apex run -e env -dir /infra/pre terraform apply
- apex deploy -e env
- apex run -e env -dir /infra/post terraform apply

Thoughts? Happy to do up the changes in a PR.

marioharper:
Took a crack at the proposed `exec` command solution for #619. Did not remove any of the current `apex infra` stuff. Just added the new `exec` command. 

Adds an `apex exec` command which wraps your command while exporting the following env variables:
- apex_environment = env
- apex_function_role = role ARN
- apex_function_<FN_NAME> = lambda ARN

Adds a `apex exec --dir <directory>` flag which specifies which directory to execute the command in. Defaults to the pwd.


cmd/apex/exec/exec.go
```diff
@@ -0,0 +1,80 @@
...
+	Short:   "Command execution passthrough",
+	Example: example,
+	RunE:    run,
+}
+
+// Initialize.
+func init() {
+	root.Register(Command)
+	f := Command.Flags()
+	f.StringVarP(&dir, "dir", "d", "", "Which directory to execute command from")
```
tj: Should be fine to omit this sine we have `-C dir`:
https://github.com/apex/apex/blob/master/cmd/apex/root/root.go#L62

cmd/apex/exec/exec.go
```diff
@@ -0,0 +1,80 @@
...
+		if err != nil {
+			return err
+		}
+
+		dir = pwd // default directory to pwd
+	}
+
+	if len(args) > 0 {
+		command = args[0]
+		args = append(args[:0], args[1:]...) // remove command from args array
```
tj: think `args = args[1:]` should be ok

cmd/apex/exec/exec.go
```diff
@@ -0,0 +1,80 @@
...
+	} else {
+		return errors.New("No command specified")
+	}
+
+	err := root.Project.LoadFunctions()
+
+	// Hack to prevent initial `apex infra apply` from failing,
+	// as we load functions to expose their ARNs.
+	if err != nil {
```
tj: I forget what this was for in `infra` hahaha.. does it work fine without this? I should add a better comment

cmd/apex/exec/exec.go
marioharper: good call 👍

cmd/apex/exec/exec.go
marioharper: I tried it with -C but -C expects your project.json config file to be in the directory you're changing to. This is because in the pre-run it changes the directory first, then opens the project. I can change that order, but I can see where that could have it's use cases.

cmd/apex/exec/exec.go
marioharper: Took it out, nothing blew up so far lol

tj:
Oh damn yeah hahah I hit that with the `infra` command, I had to inject `--` https://github.com/apex/apex/blob/master/cmd/apex/main.go#L82-L101, pretty big hack

PR #627, opened 2016-12-28
add function arn to list output

### LINKED ISSUE #626: Add function ARN to `apex list` output
Adding the lambda function ARN to `apex list` would be really useful, e.g. for quickly connecting a lambda function to an Alexa skill without having to open the AWS console

It's a 4 line change and uses an AWS API result which is otherwise thrown away. Since it seems minor, I'll submit a PR just in case it gets accepted.

ian-kent:
Adds the lambda function ARN to `apex list` output - closes #626


tj:
Thanks!

PR #634, opened 2017-01-04
improve message if curl failed due to permission error

chris-rock:
This improvements helps users to see if the installation was successful or not.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -47,8 +47,13 @@ if [ -z $LATEST ] ; then
   echo "Error requesting. Download binary from https://github.com/apex/apex/releases"
   exit 1
 else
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -47,8 +47,13 @@ if [ -z $LATEST ] ; then
   echo "Error requesting. Download binary from https://github.com/apex/apex/releases"
   exit 1
 else
-  curl -sL https://github.com/apex/apex/releases/download/$LATEST/apex_$PLATFORM -o $DEST
-  chmod +x $DEST
+  echo "Download apex binary from https://github.com/apex/apex/releases/download/$LATEST/apex_$PLATFORM to $DEST"
```
tj: Maybe "Downloading" so people don't mistake that as an action they need to take. Otherwise LGTM!

tj:
thanks!

PR #636, opened 2017-01-06
Updated runtime for Golang since AWS no longer supports nodejs 0.10

yoonian:
Just changed runtime string only not stubs(index.js, ...)

When I run

> apex deploy

It shows up today. So I just did

> InvalidParameterValueException: The runtime parameter of nodejs is no longer supported for creating or updating AWS Lambda functions. We recommend you use the new runtime (nodejs4.3) while creating or updating functions.

### DIFF
--- a/plugins/golang/golang.go
+++ b/plugins/golang/golang.go
@@ -6,14 +6,14 @@ import (
 	"github.com/apex/apex/plugins/nodejs"
 )
 
-func init() {
-	function.RegisterPlugin("golang", &Plugin{})
-}
-
 const (
 	// Runtime name used by Apex
 	Runtime = "golang"
 )
+
+func init() {
+	function.RegisterPlugin(Runtime, &Plugin{})
+}
 
 // Plugin implementation.
 type Plugin struct{}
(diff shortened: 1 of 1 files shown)


DaveBlooman:
Probably needs an immediate release, otherwise no golang lambda can be deployed

plugins/golang/golang.go
```diff
@@ -6,15 +6,15 @@ import (
 	"github.com/apex/apex/plugins/nodejs"
 )
 
-func init() {
```
mooyoul: It seems nothing changed on `init` function.

mooyoul: Why function location was moved? just curious :)

plugins/golang/golang.go
```diff
@@ -29,7 +29,7 @@ func (p *Plugin) Open(fn *function.Function) error {
 	}
 
 	fn.Shim = true
-	fn.Runtime = nodejs.Runtime
+	fn.Runtime = nodejs.Runtime43
```
mooyoul: yeah this looks good to me :)

polds:
This was addressed already in #637

PR #646, opened 2017-01-16
Don't deploy broken code

davidblurton:
Webpack returns exit code zero even if the build fails (??!) unless you pass the `--bail` flag.

The bail flag ensure that commands like deploy stop on a failing build.


tj:
Ohh damn, thanks!

davidblurton:
Yeah, bit unexpected.

Issue #364, opened 2016-04-14
Updating alias on deploy


dekz:
Thanks for your work on Apex! It's a super easy to use tool for us.

We are using multiple AWS accounts, but not exactly one per environment. Production, NonProd are split. This means NonProd might have a few environments under it, or per-person lambdas.

We are seeing an issue with deploying and having aliases in a CI pipeline.

Current State: 
- Version 3 has tags `[dev, staging]`

New code change:

```
> apex deploy --alias dev
   • config unchanged          function=failure_alerts
   • updating function         function=failure_alerts
   • updated alias dev         function=failure_alerts version=4
   • function updated          function=failure_alerts name=test_failure_alerts version=4
```

Current State: 
- Version 3 has tags `[staging]`
- Version 4 has tags `[dev]`

This passes testing and continues to the staging environment:

```
> apex deploy --alias staging
   • config unchanged          function=failure_alerts
   • code unchanged            function=failure_alerts
```

Current State: 
- Version 3 has tags `[staging]`
- Version 4 has tags `[dev]`

Expected State:
- Version 4 has tags `[dev, staging]`

dekz:
I can side step the Noop by setting fake environment variables on each deploy stage, but this creates excess lambda versions.

mthenw:
Thanks for reporting. That's right, we should update alias even though there are no changes in code or configuration.

mthenw:
I've been thinking about this for a bit and I'm not sure how we should resolve that. Possible solutions:

1. Update alias every time alias version is not the latest version.
2. Compare configuration under alias with the latest version configuration and update only if those configs differs. 

@tj any preference?

tj:
hmm if I understand correctly, sounds like we just maybe need a set of commands for alias CRUD. I think trying to add alias operations to deploy could be a little confusing. When I see those commands I don't really see a clear behaviour in mind, like oh that obviously does X.

If I run `apex deploy --alias tj` as dev-level isolation, I'd expect it to be a new release, so I'm not sure how we could have both there.

Maybe we're better off using "current" as we always have, and delegating all other alias operations to `apex alias`, but then we need more CRUD to grab versions etc. It would be more like `git tag` though, which could feel more intuitive.

mthenw:
yeah, I was also think about what we have now as a "correct" but on the other hand would be cool to support mentioned use case. 

@dekz I'm removing this issue from upcoming release. Let's wait for more feedback.

joshrtay:
[...]

PR #647, opened 2017-01-19
Add alias command

llun:
Add alias command base on this discussion #364 but I change from the last comment a bit by adding function name as the last argument instead of version name

- apex alias <alias name> for adding alias all functions with the name
- apex alias <alias name> <function name> for adding alias on specific function
- apex alias -v <version> <alias name> for adding alias on specific version
- apex alias -v <version> <alias name> <function name> for adding alias on specific version and specific function

### DIFF
--- a/cmd/apex/alias/alias.go
+++ b/cmd/apex/alias/alias.go
@@ -1,0 +1,72 @@
+package alias
+
+import (
+	"errors"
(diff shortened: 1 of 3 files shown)


tj:
Awesome! Thanks :D

PR #652, opened 2017-01-22
[WIP] KMS Env Vars

DaveBlooman:
Initial idea for KMS support.  Need to move some things around and handle errors correctly.

function.json would look like 

```json
{
  "name": "foobar",
  "description": "foo",
  "memory": 128,
  "timeout": 5,
  "runtime": "nodejs4.3",
  "role": "arn:aws:iam::00000000000:role/lambda_function",
  "environment": {
    "foo": "someAPIkey"
  },
  "encrypted_vars": true,
  "key_id": "c2373eb5-166e-4c10-9ef0-b40d6036a9c4"
}

```

What that looks like in the console.  Not sure if we want the names encrypted, this would be where having a bool on each variable would come in nice, though if you want everything encrypted, the current approach is better.

screenshot 2017-01-22 00 41 37


tj:
Awesome I'll try and check this out tonight!

function/function.go
```diff
@@ -872,3 +916,42 @@ func environ(env map[string]*string) []string {
 
 	return pairs
 }
+
+func decryptEnvs(env map[string]*string, service *Function) ([]string, error) {
```
tj: Do you know if Lambda will decrypt these for us? I know it does with the default KMS setup, but if we're encrypting before sending anything over the wire, does it double-encrypt in that case? It would be nice if we didn't have to add any Lambda-side code for decrypting

DaveBlooman: @tj No, it will not decrypt for us. I guess the use case here is highly sensitive variables that should never be visible a plain text form.  I was thinking of writing a wrapper for the Go version that decrypts for you, for the other languages, AWS provides a way to get decrypt during the lambda.  I doubt that this would be used for API gateway for example, but we have non time sensitive lambdas that use an RDS password

PR #653, opened 2017-01-23
Add alias on another alias

llun:
Add option for alias on another alias, make alias command easier to use. Also, change default alias version to `current` instead of `$LATEST` which's moving version.


tj:
Great! Thanks

PR #671, opened 2017-02-07
Fix regex for older grep versions

wujiang:
Fix the install script on OpenBSD which has an older version of grep.

See comment here https://github.com/apex/apex/issues/400#issuecomment-278069309


tj:
haha damn, so much for this being a convenient script, thanks!

PR #672, opened 2017-02-08
Added an example of using python 3

jacobrvh:
This is an example of how to use apex with python 3. There were are no actual code changes to get this working, however I think with this could be incorporated into apex to have native python 3 support just like go.

The import things to get this working were making an executable file named main since that is what the shim looks for and setting the shebang to use python3 with the unbuffered parameter to allow the shim to read from stdout immediately. I also redirected normal stdout to stderr since the shim is expecting stdout to be the result.

I plan on supplying another example with dependencies and another example with dependencies built against the linux ami using docker.


codingjoe:
@jacobrvh thanks for writing all the example code. I use docker to compile python dependencies too. I think that is a very good approach and should be documented.
The approach to add Python 3 support is a hack ;) I'd prefer to see a proper integration. There should be actually much missing. Its only a minor API change to my opinion.

I'm not a gopher too, but I think I should be only adding another file like this:
https://github.com/apex/apex/blob/e5fff82ba2969169e13ca7d5f1c65f1f67b0456e/plugins/python/python.go

A little side note tho:
I think that the name python should reference python 3.6 not python 2.7. Since python 2 is a legacy Version and will be finally discontinued in 3 years.

_examples/python3/project.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "python3",
+  "description": "Python 3 example project",
+  "role": "arn:aws:iam::293503197324:role/apex_lambda_function",
+  "runtime": "nodejs4.3",
```
codingjoe: It should be:
```json
{
  "runtime": "python3.6",
  "handler": "main.handle"
}
```

skabbes:
Cool technique, yes.  I actually added an example of how to run python 3 code here:

https://github.com/apex/apex/pull/813

codingjoe:
@skabbes cool! I think it might still make sense to add an example for how to use Docker. Especially when you want to use Python packages with C dependencies or C libraries with Python bindings it is very useful to have Docker in place to create the deployment package.

In fact, I would be ever better if apex would always just use Docker to create the deployment bundle, but that would be a bit of an overkill ;)

jacobrvh:
This code is no longer that helpful now that lambda supports python 3.

PR #676, opened 2017-02-13
infra: prepend apex args so user are appended more naturally

### LINKED ISSUE #434: apex infra: "The apply command expects at most one argument."
```
$ apex infra -e dev apply terraform.plan
The apply command expects at most one argument.
Usage: terraform apply [options] [DIR-OR-PLAN]

  Builds or changes infrastructure according to Terraform configuration
  files in DIR.

   ...
```

Is apex passing this positional argument to terraform for me already? 

This seems possibly related to https://github.com/apex/apex/issues/421, since it involves the parsing of apex/terraform command line arguments.

---------

```
$ apex version
Apex version 0.9.0
$ terraform version
Terraform v0.6.16
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.11.3
BuildVersion:   15D21
```

dominicbarnes:
Fixes #434 by prepending "system" (ie: apex) args before invoking terraform. This behavior feels more natural, as arguments specified by the user should fall last in a case like this, where apex is basically acting as an alias.


dominicbarnes:
fwiw, this should have no effect on existing behavior, except to fix this particular use-case.

tj:
sweet, I barely use TF now hahah, I completely forget why it's sensitive to the ordering for -var's, but based on #434 wasn't it an issue with `plan` as well? Maybe this was fixed:

```
You can't set variables with the '-var' or '-var-file' flag
when you're applying a plan file
```

dominicbarnes:
Yeah, #434 specifically mentioned `apply` with a `.plan` file as an argument. When I use `plan` however, I get the "The plan command expects at most one argument." error.

I hadn't considered the "You can't set variables" in my fix though, so I'm not sure if this will fix. Maybe there needs to be another flag that tells apex not to inject variables? (starting to get really janky, so perhaps we should see if TF has changed anything in this regard since a year ago)

tj:
I haven't touched terraform in months now haha.. I'll have to give this some more attention, I forget the current state of things

hsienchiaolee:
@tj Any updates on this? Having trouble because `apply` with `.plan` files won't work at the moment. It throws `Too many command line arguments. Configuration path expected` error (Terraform v0.11.7, Apex version 1.0.0-rc2). Alternatively, can we skip `functionVars` when a `.plan` file is provided?

hsienchiaolee:
@tj You are right, this won't fix the problem I am experiencing. The problem that I was having is that when using `terraform apply terraform.plan` with the plan file generated through `terraform plan -out=terraform.plan`, it complains when additional arguments are passed in after the plan file:
```
Too many command line arguments. Configuration path expected
```
When reversing the order and supplying `-var` options first (`terraform apply -var foo=bar terraform.plan`), it would also throw an error:
```
You can't set variables with the '-var' or '-var-file' flag
when you're applying a plan file. The variables used when
the plan was created will be used. If you wish to use different
variable values, create a new plan file.
```

atistler:
Same issue here, can't apply a plan file.

```
→ apex infra plan -out tfplan
→ apex infra apply tfplan
Too many command line arguments. Configuration path expected.
   ⨯ Error: exit status 1
```

PR #692, opened 2017-02-27
update rust example's dependency to use rust-apex 0.2.0

doomsplayer:
I've update rust-apex to 0.2.0 so an update is also needed here.


tj:
thanks!

Issue #698, opened 2017-03-06
Feature request: support Step Functions (State Machines)


temoto:
It feels appropriate in apex since StepFunctions and Lambda are so closely related.

Current best way of uploading state machines using CloudFormation has downsides (just sucks) of speed and error in state code breaks CF stack forever, so you have to delete it before retry.

Sorry, I understand it's miserable to just request without providing code, consider this more of a question/poll. Patch is coming maybe, can't promise.

temoto:
These two patches enabled `errorType` passing for me. Which was the only critical blocker.
https://github.com/apex/go-apex/pull/48
https://github.com/apex/apex/pull/699

The rest of this feature is probably around deployment/logs etc.

nedlowe:
This would be really awesome - the current options for Step Functions are pretty poor.

PR #699, opened 2017-03-08
shim: support errorType and stackTrace

temoto:
Required for StepFunctions support
https://github.com/apex/apex/issues/698


temoto:
ping

PR #705, opened 2017-03-17
Fix formatting in docs

AndrewKvalheim:
Looks like some unusual whitespace was preventing the Markdown heading from rendering.

Apologies if there's anything amiss with this PR... I'm trying out GitHub's web interface for "editing" files.


tj:
weird, thanks!

PR #712, opened 2017-03-25
Update function deploy to update the specified alias even if code has not changed.

### LINKED ISSUE #686: Apply --alias even if code is unchanged
I was wondering if you could consider taking a PR for this use case. Let me describe what I'm doing first and maybe you can see an obvious solution that I'm derping on:

- We have a set of lambda functions running our application.
- We have multiple environments, each with their own set of dbs/infrastructure etc.
- We have partitioned lambda functions via aliases. So theres a 'production' alias for production, 'staging' for staging, and individually named ones for each developer's environment.
- The event source for the lambda functions are through API gateway. They are separated by different stages in API Gateway, with each staging pointing to their respective alias.
- The problem comes in trying to setup a CD pipeline to bring an environment to a version.
    - We could use `apex deploy --alias {environment}`, but an alias is not deployed if the code is unchanged. This means that, in the scenario that you're deploying to production after a successful staging deploy, production won't move since the latest lambda version is the same.
    - We were thinking about using `apex alias {environment}` but that takes the environment to the latest code, which is not what we want, since each developer could be deploying their own stuff, pushing the latest version forward.
    - We also considered just having a different lambda function per environment, but that doesn't seem to really gel with what AWS recommend in their docs: http://docs.aws.amazon.com/lambda/latest/dg/versioning-aliases.html

What I want to do is change the `apex deploy` to still move the alias if the latest version of the lambda function is the same as the code being deployed. What do you think? Am I missing something obvious?

clark-pan:
Closes #686


clark-pan:
@tj bump

tj:
hmm oddly seems to not add the alias, no error from the UpdateAlias call though haha

tj:
ah weird the API / Console just seem really delayed

tj:
we'll have to check for an error in `Function.removeVersions()` if the func has an alias as well, forgot to do that

PR #719, opened 2017-04-01
refactor example to use nodejs version 6.10

thewillhuang:
updates babel2 webpack example to use nodejs6.10


alecl:
Do we still need the babel transpiler given nodejs 6.1 is more es6 compliant natively?

Can this webpack config be modified to not use the babel part? https://github.com/apex/apex/blob/master/_examples/babel-webpack2/webpack.config.babel.js

PR #732, opened 2017-04-26
Add documentation for setting alias to another alias

johnlinvc:
Add documentation for setting a alias to another alias.
Common use case is for promoting between multiple stages.


tj:
thanks!

tj:
although won't this point to the alias and not the version it represents? so if "dev" were updated it would no longer be correct

PR #735, opened 2017-04-26
Ensure Function.GetVersionFromAlias error is shown. Fixes #734

### LINKED ISSUE #734: Ensure Function.GetVersionFromAlias() error is shown
- __Max OS X 10.11.6__:
  - __Apex version 0.14.0__:

While investigating #728, I found the error to be very misleading. It turned out that the error was a red herring. I was getting:
```
ValidationException: 1 validation error detected: Value 'current' at 'functionVersion' failed to satisfy constraint: Member must satisfy regular expression pattern: (\$LATEST|[0-9]+)
	status code: 400, request id: xxxxx
```

But really, it was an auth error:
```
AccessDeniedException: User: arn:aws:iam::xxxxx:user/apex is not authorized to perform: lambda:ListAliases on resource: arn:aws:lambda:us-east-1:xxxx:function:xxxxx
	status code: 403, request id: xxxx function=xxxxx
```

It turned out that the auth error returned from `GetVersionFromAlias()` was being overwritten by the error returned from `CreateOrUpdateAlias()` in Project.go:
```
version, err := fn.GetVersionFromAlias(version)
if err != nil {
    err = fmt.Errorf("function %s: %s", fn.Name, err)
}

err = fn.CreateOrUpdateAlias(alias, version)
if err != nil {
    // this is overwriting the previous error
    err = fmt.Errorf("function %s: %s", fn.Name, err)
}
```

noelmansour:
If GetVersionFromAlias() returns an error which subsequently results in
an error in CreateOrUpdateAlias(), the local err variable will be
overwritten with the CreateOrUpdateAlias() error.

Instead, add both instances to the errs channel.

### DIFF
--- a/project/project.go
+++ b/project/project.go
@@ -390,14 +390,14 @@ func (p *Project) CreateOrUpdateAlias(alias, version string) error {
 				version, err := fn.GetVersionFromAlias(version)
 				if err != nil {
 					err = fmt.Errorf("function %s: %s", fn.Name, err)
+					errs <- err
 				}
 
 				err = fn.CreateOrUpdateAlias(alias, version)
 				if err != nil {
 					err = fmt.Errorf("function %s: %s", fn.Name, err)
+					errs <- err
 				}
-
-				errs <- err
 			}()
 		}
 


noelmansour:
Fixes #734

tj:
oh I think we're missing a `return` in there, I'd rather have that than pass through to the CreateOrUpdateAlias() call

noelmansour:
Done.

Also, the nil check in the for errs for loop can probably be removed now. I can do that as well if you'd like.

tj:
sgtm!

project/project.go
```diff
@@ -406,9 +407,7 @@ func (p *Project) CreateOrUpdateAlias(alias, version string) error {
 	}()
 
 	for err := range errs {
-		if err != nil {
-			return err
-		}
+		return err
```
tj: I think we could refactor this to just be `errC := make(error, len(p.Functions))` and `return <-errC` but this should be fine

project/project.go
```diff
@@ -391,7 +391,7 @@ func (p *Project) CreateOrUpdateAlias(alias, version string) error {
 				if err != nil {
 					err = fmt.Errorf("function %s: %s", fn.Name, err)
 					errs <- err
-					return;
+					return
```
noelmansour: Hehe whoops. Blame it on the Java I've been writing.

PR #744, opened 2017-05-04
Pass maps that describe all apex functions to terraform Closes #737

### LINKED ISSUE #737: Proposal for terraform variable exports
Right now, Apex exports the functions' arn into a separate variable for each function. As a consequence, a newly created function will have to be included manually into terraform. This also means that the files will be valid terraform files if we do an `apex infra plan` into the infrastructure if we forget to remove it as a variable.

A solution would be to export the function names as a map to terraform. So on terraform's scripts it would become:
```
variable "apex_functions" {
    type: "map"
}

...

// Replace the following with a valid resource that needs the lambda's arn
resource "someresource" "somename" {
   lambda_function = "${var.apex_functions["somefunction"]}"
}
```
The above will fail on plan if the function is not defined, and we never need to explicitely import the function as it will become immediately available after running `apex deploy`.

alialamine:
Hello,

This will allow apex to pass all the functions in a map to terraform.


tj:
ah cool, didn't know TF allowed that. Maybe we should break some of that serialization out into separate funcs to tidy up a bit

alialamine:
This sould allow for an easy addition of other variables without increasing the complexity of the code (just add another function and execute it)

I also separated them into another file as I'm planning to do some work on the infra command as I see it's very limiting and it has the possibilities of it automating a lot of things in the infrastructure (will open issues for each as I get the need for them)

docs/infra.md
```diff
@@ -32,6 +32,8 @@ Currently the following variables are exposed to Terraform:
 - `apex_function_role` the Lambda role ARN
 - `apex_function_NAME` Lambda function ARNs by __NAME__
 - `apex_function_NAME_name` Lambda function names by __NAME__
+- `apex_functions` A map of all lambda functions
```
tj: what do you think about calling it `apex_function_arns` and `apex_function_names`? then maybe we can undocument what I had before to phase those out

tj:
awesome thanks!

Issue #761, opened 2017-05-25
Installing via install.sh doesn't work


xuru:
Stats:
  - __Operating system__: MacOS Sierra version 10.12.4
  - __Apex Version__: v0.14.0

Following the installation instructions [here](http://apex.run/#installation) doesn't actually install apex.

```
$ curl https://raw.githubusercontent.com/apex/apex/master/install.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1905  100  1905    0     0   7922      0 --:--:-- --:--:-- --:--:--  7937
Downloading Apex binary from https://github.com/apex/apex/releases/download/v0.14.0zipball_url/apex_darwin_amd64 to /usr/local/bin/apex
Apex installation was successful

$ apex
/usr/local/bin/apex: line 1: Not: command not found
```

Looking at the script, I found a few things:
```
$ LATEST=$(curl -s https://api.github.com/repos/apex/apex/tags | grep -Eo '"name":.*[^\\]",'  | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)

$ echo $LATEST
v0.14.0zipball_url
```
Which ends up trying to download apex from the following url:
`https://github.com/apex/apex/releases/download/v0.14.0zipball_url/apex_darwin_amd64`
When it should be:
`https://github.com/apex/apex/releases/download/v0.14.0/apex_darwin_amd64`

I'm not good enough at all the `sed`/`cut`/etc or I would put in a pull request...

indiejoseph:
+1

Liath:
Could I get a Mac user to check that the following works: 
```curl -s https://api.github.com/repos/apex/apex/tags | grep name | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2```
At the moment it should yield `v0.14.0`

polds:
@Liath You are correct.

mrichman:
Correct.

tj:
merged the PR

PR #762, opened 2017-06-06
Remove uneeded grep parameters that are breaking BSD grep in install.sh

Liath:
Per [issue/761](https://github.com/apex/apex/issues/761). Change the grep parameters in install.sh that were breaking on OSX (and presumably for anyone else who doesn't have the privilege of GNU grep)


tj:
thanks!

PR #766, opened 2017-06-14
move -C chdir flag check to start of preparation function.

### LINKED ISSUE #757: Fix -C chdir flag
Last I checked this is too low in the execution to work properly, thus it will not pick up the functions properly etc

ianrtracey:
move -C chdir flag check to start of preparation function in order to avoid interference with other flags when using chdir. Closes #757.

celebrate

### DIFF
--- a/cmd/apex/root/root.go
+++ b/cmd/apex/root/root.go
@@ -104,6 +104,12 @@ func Prepare(c *cobra.Command, args []string) error {
 	// config defaults
 	Config = aws.NewConfig()
 
+	if chdir != "" {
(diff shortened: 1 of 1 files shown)


ianrtracey:
Not sure if this block should be before or after the new config function at:
https://github.com/apex/apex/blob/master/cmd/apex/root/root.go#L105

DavidTPate:
@ianrtracey Could you please add a test for this specific case to our test suite? That way we can make sure it works as expected and continues to work as expected.

tj:
might be a little tricky to test

PR #771, opened 2017-06-20
change exact runtime matching to prefix matching

### LINKED ISSUE #731: Remove over-engineered runtime checks in plugins
Like e5fff82ba2969169e13ca7d5f1c65f1f67b0456e, the others should be similar, right now it's annoying having to manually add versions.

ianrtracey:
- add prefix runtime checks for node & rust
- add default node version for inferencing on plugin init

Closes #731.

### DIFF
--- a/plugins/nodejs/nodejs.go
+++ b/plugins/nodejs/nodejs.go
@@ -1,27 +1,18 @@
 // Package nodejs implements the "nodejs" runtime.
 package nodejs
 
-import "github.com/apex/apex/function"
+import (
+	"github.com/apex/apex/function"
+	"strings"
+)
 
 const (
 	// Runtime name used by Apex and by AWS Lambda for Node.js 0.10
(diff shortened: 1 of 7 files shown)


plugins/nodejs/nodejs.go
```diff
@@ -19,27 +22,24 @@ const (
 
 func init() {
 	function.RegisterPlugin(Runtime, &Plugin{})
-	function.RegisterPlugin(Runtime43, &Plugin{})
```
tj: we should be good to remove the constants now (other than `Runtime`)

ianrtracey: Removed the other node Runtime constants. Had to end up changing the interference runtime as well as the golang and rust shims. Hope that was expected.

tj:
sweet LGTM! Thanks man

PR #780, opened 2017-07-04
doc/installation: DEST can be used to specify apex destination

kyanh-huynh:
doc/installation: DEST can be used to specify apex destination


tj:
Thanks!

PR #791, opened 2017-07-17
Added automatted tagging for lambdas based on the config file

### LINKED ISSUE #790: Support for adding lambda tags through json file
Currently apex does not support automatically adding or updating lambda tags read through the project.json file. I have added this support on a branch in my repo and would like to discuss this issue before opening a PR. Is there a reason this is not currently supported, and would this feature be desirable?

lucas-rudd:
Closes #790

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -703,7 +703,7 @@ func (f *Function) GroupName() string {
 
 // GetVersionFromAlias returns function version from alias name, if alias not found, return the input
 func (f *Function) GetVersionFromAlias(alias string) (string, error) {
-	var version = alias
+	version := alias
 	aliases, err := f.GetAliases()
 	if err != nil {
 		return version, err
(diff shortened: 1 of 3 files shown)


lucas-rudd:
The two changes made to the error texts was a recommendation taken from golint. The methods were not originally written by me.

tj:
sweet looks good to me other than a few tiny changes, thanks!

jonathanhughes:
@tj Could you elaborate on what changes need to be made?  We'd be happy to make the updates.  Thanks!

tj:
@jonathanhughes oops should show the review comments now, forgot to submit it

function/function.go
```diff
@@ -915,19 +928,29 @@ func (f *Function) environment() *lambda.Environment {
 	return &lambda.Environment{Variables: env}
 }
 
-// environment sorted and joined.
-func environ(env map[string]*string) []string {
+// tags for lambda calls.
+func (f *Function) tags(config *lambda.GetFunctionOutput) *lambda.TagResourceInput {
+	tags := make(map[string]*string)
```
lucas-rudd: How is that?

tj:
oops sorry I did a bad job reviewing hahah.. commenting inline

function/function.go
```diff
@@ -312,6 +313,10 @@ func (f *Function) DeployConfigAndCode(zip []byte) error {
 			SubnetIds:        aws.StringSlice(f.VPC.Subnets),
 		},
 	}
+	config, _ := f.GetConfig()
```
tj: I think we should just take the tags from the JSON config. I'd prefer not to make multiple additional calls to `GetConfig()` since it'll add more latency

lucas-rudd: We are taking the tags from the JSON config. The GetConfig call is to get the function ARN for the lambda we are intending to update. See the 'tags' method.

vaishalioswal:
Hi!How to add tags to aws lambda function using apex CLI?
Do I have to edit project config file?If yes,then what to add?
I am relatively newbie and I have tried everything..(including editing config file)

lucas-rudd:
Hi @vaishalioswal! Unfortunately, something went awry with the release that this was scheduled to go in, and Apex was rolled backed to a previous version not containing these changes.

See issue #790 for more information.

Tags are added in the config file for either the lambda function in question, or the base project config.

An example of how this would work upon its implementation is as such.

```
{
  "description": "Node.js example function",
  "runtime": "nodejs",
  "memory": 128,
  "timeout": 5,
  "role": "arn:aws:iam::293503197324:role/lambda"
  "tags": {
    "foo": "bar"
  }
}
```

PR #796, opened 2017-07-31
shim: Support Binary data in Node.js shim

bad79s:
The api gateway supports binary.
However, you can not handle binary using go-apex.
So we added the identifier "\_\_BINARY\_\_" to the Node.js Shim for go-apex.

https://github.com/apex/go-apex/issues/58


tj:
hmm I've had no issues with binary in the API Gateway proxy stuff

bad79s:
Currently, you can do this using the proxy function of API Gateway.
But I want to be able to use it without using a proxy.
So I made it.

PR #804, opened 2017-08-20
fix to add secret.json and move role from project.json. Closes #628

### LINKED ISSUE #628: Store role outside project.json
Is there a way to store the role value outside of `project.json`?

The current setup means my AWS account ID is part of project.json, which complicates having a public repo on GitHub where the lambda function is already deployed.

178inaba:
This pull request moves the role from `project.json` to `secret.json` to make it easy to publish the application.
`secret.json` is hidden from git by `.gitignore`.


178inaba:
Error will be fixed with the following pull request.
https://github.com/apex/apex/pull/803

It is not an error caused by this pull request.

PR #812, opened 2017-09-07
Update the handler

jpmonette:
Updated the `handler` value. Previous value was returning this when invoking the functions:

` ⨯ Error: function response: Handler 'default' missing on module 'lib'`


vespertilian:
@jpmonette Does this work fo you? 

For me it seems to need to be

`"handler": "lib/index.default"`

not `index.handle` 

As the functions are exported as default.

PR #813, opened 2017-09-12
examples: add example python3.6 function

skabbes:
While debugging up, it wasn't 100% clear to me that python 3.6 was working.  I created a test / example to verify that and figured I'd at least share it here.


tj:
thanks!

Issue #380, opened 2016-04-23
Hooks don't work on Windows


niba:
I'm trying to run an example babel-webpack without success. When I run `apex build requester` I'm getting `Error: build hook:` Here is a full output with log level = debug
```
   • loading functions in functions
   • loading function in functions\requester
   • creating build            function=requester
   • hook                      command=../../node_modules/.bin/webpack --config ../../webpack.config.js function=requester hook=build
   ⨯ Error: build hook:

```
I tried to change build hook to something different but even the simplest command (like `echo`) doesn't work

tj:
hmm I'm guessing that command fails from the path separators not being \ on windows, but if `echo` fails too I'm not sure hmm, I can't help debug windows unfortunately :( other than guessing, no vm set up at the moment

monamu:
Hi.
I also encountered the same situation . The message was as follows .

`couldn't read role from infrastructure: exec: "sh": executable file not found in %PATH%`

mthenw:
https://github.com/apex/apex/blob/b15a27724e55610a7240c3aa0c708c5127c48078/infra/infra.go#L80

tj:
hmm I don't recall why we're using shells anywhere haha.. we should be able to swap all those with regular execs

mooyoul:
@tj Actually, There are no histories about exec without `sh` command. haha

Anyway i'll take look about this issue with my windows machine on this weekend :)

enzo2:
Would like to use Go on Windows, is there a way around this?

Liath:
@enzo2 and others looking for a workaround.
Because apex is trying to exec `sh` you can place a file called `sh.bat` in the project folder with something like this in it:
```
set args=%*
cmd /c %args:~3%
```
(The `%:~3%` bit will trim off the '-c' from `hooks/hooks.go:`.)

I then also set the hooks like this:
```
    "hooks": {
		"build": "build",
		"clean": "clean"
    },
```
And added a build.bat and clean.bat for Windows and bash scripts named `build` and `clean` for linux.

shaishab:
I am also facing same problem from windows and tried @Liath solution but no luck 

used like:-

"hooks": {
    "build": "build",
    "clean": "clean"
  }   

any help please

Liath:
As a follow-up to my last post for people waiting on the above PR, try putting sh.bat in the function folder too. Seems like weird path shenanigans were making it work for me originally.

PR #814, opened 2017-09-27
Fix for Windows hooks

Liath:
See issue #380

### DIFF
--- a/plugins/hooks/hooks.go
+++ b/plugins/hooks/hooks.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


Liath:
`plugins/hooks/hooks.go:70` wasn't producing any output because `b` is empty and the error message is in `err` instead in this case. Idk what the correct way to handle this is, but it would be nice to print `err` instead of `b` when it has something useful to say.

Liath:
And maybe we should do `powershell -Command` instead of `cmd /c`. Powershell is basically everywhere and has aliases for most nix commands, `mv`, `rm`, ls`, etc…

tj:
sounds reasonable to me!

PR #822, opened 2017-10-09
shim: allow for concurrent calls to handler

skabbes:
After talking with a few users of Up and Apex, the correctness of the shim is often coming into question.  I am not saying that it is not correct currently, but at the very least the correctness is very tied to implementation details of Lambda.

This PR is tightly coupled to [go-apex#60](https://github.com/apex/go-apex/pull/60) and allows for concurrent execution of lambda handlers without any hesitations about the correctness.

A separate benefit is that integrating this will dramatically reduce the surface area of exploration for [this gnarly bug](https://github.com/apex/up/issues/276).

If performance is an issue, I can refactor this to not handle concurrency, but to only simply assert on our assumptions of lambda.


skabbes:
@berzniz here is another change which would facilitate our debugging

tj:
thanks man!

Issue #59, opened 2016-01-08
Define configuration shape and semantics


tj:
## Problem:

- We haven't fully stabilized or future-proofed the config
- Function defaults are currently top-level properties

## Cases to consider:

- [ ] environment variables
- [ ] API gateway
- [ ] input sources
- [ ] function-level config (not just environment variables)
- [ ] project-level defaults and/or for all of the above
- [ ] environments specific config
  - tuning of function config (role, timeout, etc)
  - environment variables for functions

## Thoughts 

We could tackle this a few ways. One would be to just keep throwing config into package.json or function.json respectively. With this we have to use a lot of nesting for clarity.

Another would be to treat gateways and others as first-class things like functions, using their own dirs.

Another alternative might be to have a utility which basically lets you decompose project.json into directories of the same key names, as you wish. For example:

```json

{
  "functions": {
    "defaults": {
      "runtime": "node",
      "memory": 128
    }
  },
  "gateways": {
    "api": {
      "/pets/{name}": {}
    }
  }
}
```

Could be:

```
./config
  ./functions
    ./defaults.json
  ./gateways
    ./api.json
```

Or:

```
./config
  ./functions.json
  ./gateways.json
```

You can expand any object to dirs as-desired, while maintaining a well-known shape. I'd like a unifying concept like this if we can pull it off without hurting UX.

mthenw:
> function-level environment variables that are auto-populated via process.env and os.Setenv() etc

not sure how to achieve it now, this function is not used anywhere https://github.com/apex/apex/blob/master/function/function.go#L104

> I'm generally not a fan of environment variables since they're pretty restrictive shape-wise, so I'd definitely prefer if those were not the primary configuration mechanism.

Heh, I think this should be default configuration mechanism because it's simple. 

I thinking about a object in `project.json` or `function.json` under `env` key. And of course --env support in `apex deploy`. That's all.

tj:
I guess the problem I have with envs here is just that they're not great at describing more complex config. The use-case I have in mind is log-collector which will support dozens of providers each with a variety of config. Env is decent for setting db endpoints etc but beyond a few values it's a little crazy (unless we just assign json to the env var and let the user parse, that could work too). 

AWS ECS does similar with docker config, you just do `ECS_ENGINE_AUTH_DATA={username,password,blahblah}`, so I suppose we could do that with "env":

```json
{
  "env" {
    "ECS_ENGINE_AUTH_DATA": { "username" ... }
  }
}
```

[...]

PR #833, opened 2017-10-26
resolves go-apex #59 - adds support to shim for custom error types to be returned by go lambda functions to support AWS Step Function retries

savaki:
Hopefully this will fall into the category of trivial change.  Currently the go shim does not allow custom error types to be returned.  Consequently, the retry feature of AWS Step Functions is impractical to use.  This change along with the corresponding go-apex change would allow go code to return custom error types.

http://docs.aws.amazon.com/step-functions/latest/dg/tutorial-handling-error-conditions.html


Liath:
I felt a compulsion to unpack the changed lines from `shim/bindata.go`. Here's a js one-liner for the curious.

```javascript
zlib.gunzipSync(Buffer.from('<bindata>'.split('').map(x => x.charCodeAt()))).toString()
```

PR #851, opened 2017-12-11
update rust example

doomsplayer:
Update `rust-apex` to 0.3 which brings in version 1.0 of `serde` as well as start using `failure` crate to do error handling in rust example.


tj:
nice! looks a lot simpler

PR #859, opened 2018-01-09
added auto dependency install on nodejs

ngoyal16:
hook to auto install npm packages


Liath:
So this is an example of setting build hooks in function.json?

ngoyal16:
yes

PR #868, opened 2018-02-06
project.json_sub with wrong handler

nerijunior:
Using handler as "lib/index.handle" lambda returns error because the handler don't exists


tj:
weird hahah I don't know how the name `json_sub` got in there either, seems... incorrect haha

nerijunior:
@tj hahahah my bad. Tks for the merge.

PR #875, opened 2018-04-02
add AWS endpoint flag. Closes #794

### LINKED ISSUE #794: Support for altassian localstack
Is it possibly to customize the AWS endpoints' URLs that apex uses and make it deploy functions on an altassian localstack (https://github.com/localstack/localstack) instead of the real AWS service?

I couldn't find a way of doing that with the available config options, but if that's possible I think it would be a very interesting setup for developers because it provides a way of doing local testing of the lambda functions without an active connection to AWS, and will also avoid paying AWS fees while developing.

jharriman:
Adds a flag for setting the endpoint used by the AWS Go SDK. 

An endpoint is defined in the SDK as follows:
```
// An optional endpoint URL (hostname only or fully qualified URI)
// that overrides the default generated endpoint for a client. Set this
// to `""` to use the default generated endpoint.
//
// @note You must still provide a `Region` value when specifying an
//   endpoint for a client.
Endpoint *string
```
Setting it has the effect of directing all AWS API calls to the provided endpoint: e.g. if you set it to http://localhost:4574 while running [localstack](https://github.com/localstack/localstack), all Lambda API calls will be directed to your `localstack` instance instead of Amazon.

In addition to allowing `apex` to hit `localstack` (#794), this change allows `apex` to call _any_ custom Lambda API implementation.

Note: Unfortunately both `e` and `E` are already in use as shorthand variables, so I didn't add a shorthand here. If you can recommend a shorthand flag character, I will happily update the PR.


jharriman:
@tj Sorry to ping, but this is exactly a 7 line change (ignoring whitespace) that would enable the use of apex in _several_ different kinds of offline environments. Anything I can do to get this merged?

tj:
LGTM, sorry for the delay, hard to find time for OSS these days

jharriman:
@tj Thanks! I understand. Best of luck with your current project!

PR #878, opened 2018-04-12
Add dependency for go example

pratik60:
I don't think the example should assume dependencies are installed. This step explicitly adds this in the example. 

The python example is very similar and installs dependencies too.


tj:
I guess it doesn't realllly matter since it's just an example hahah, people can remove it if they want. Thanks!

Issue #835, opened 2017-10-26
Use Assume_role in AWS profile


robh007:
If you have a bug report please let us know the following:

  - __Operating system__: macOS Sierra 10.12.6
  - __Apex Version__: 0.16.0

Apex doesn't appear to honour assume_role when using profiles with AWS CLI

The AWS cli allows you to configure access to different AWS accounts using profiles. These profiles rather than containing keys can have an assume_role parameter. This means that if you are using cross account roles from a single Authentication account into a sub account AWS CLI takes care of gathering the API keys and Session token etc. 

Link to AWS docs. 
http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html

So I have a default profile configured which gives me the ability to assume role into a sub account.

when issuing an apex init command I get the following error.

**apex init -e apex-dev -p apex-profile -r us-east-1 --log-level debug**

**Error: creating role: NoCredentialProviders: no valid providers in chain. Deprecated.
        For verbose messaging see aws.Config.CredentialsChainVerboseErrors**

The same goes if I issue

**apex init -e apex-dev -r us-east-1 --log-level debug**

and export AWS_PROFILE=apex-dev

However if I use AWS CLI I can lookup AWS resources with no issues.

_Example AWS CLI Configuration

~/.aws/credentials

[default]
aws_access_key_id = ACCESSKEYID
aws_secret_access_key=MYSECRETACCESSKEY

~/.aws/config
[resource_accounta]
assume_role = arn:aws:iam::account_number:role/accounta_role
source_profile = default
region = eu-west-1

timmersuk:
I've not tested this, but you may be able to get the current release of apex to support use of your shared config by setting the env variable AWS_SDK_LOAD_CONFIG to a truthy value (cheesy wording AWSes not mine - see https://docs.aws.amazon.com/sdk-for-go/api/aws/session/).

Changes in the referenced fork switch this on in the apex code, however I had further issues as I also needed to support use of MFA tokens

PR #886, opened 2018-05-19
AWS SDK updates

jrnt30:
I recently ran into an issue attempting to use Role Assumption with Apex #835.  

Looking around there were a few other Apex related issues indicating that it's use should be supported with the `AWS_SDK_LOAD_CONFIG=true`.  There was [a regression](https://github.com/aws/aws-sdk-go/pull/1770) in the underlying AWS SDK, updating to the latest SDK resolves this issue.  

NOTE: Users still need to provide `AWS_SDK_LOAD_CONFIG=true` to properly load the region information and assume the roles, but I have tested with both the `--profile` option and environment support successfully .

- Updating the pinned AWS SDK version to support role assumption properly
- `dep ensure` for other deps not covered here
- Updated Makefile to allow for quick installation of `make build` target requirements
- Added sample Goreleaser config to allow for injection of git tag

NOTE: Wasn't sure what the CI/CD or build process was and didn't want to
complicate things.  Happy to either include or remove the entry


tj:
awesome thanks!

Issue #887, opened 2018-05-27
Can't deploy clojure example using clojure 1.9.0


stevenarvar:
From https://github.com/apex/apex, I tried to deploy and run _examples/clojure.  Ran fine as expect, but when I change the project.clj to use clojure 1.9.0, I get this Error: function response: Error loading class org.example.lambsay

  - __Operating system__: Max OS 10.13.4
  - __Apex Version__:  1.0.0-rc2

>> java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

>> apex deploy -ldebug
   • loading functions in functions
   • open                      env= function=say
   • deploying 1 functions    
   • deploying                 env= function=say
   • building zip              env= function=say
   • creating build            env= function=say
   • hook                      command=lein uberjar && mv target/*-standalone.jar target/apex.jar env= function=say hook=build
   • appending compiled files  env= function=say
   • created build (5.6 MB)    env= function=say
   • fetching config           env= function=say
   • config unchanged          env= function=say
   • code changed              env= function=say local=/QzCFbx+k7x/bqCwBDk16yTHcGqEZYqi9DBrc3hRS6s= remote=Koo+N8BfH2XxXn4PZTs0XaE4E1EbwkNz3f+icIXU08o=
   • updating function         env= function=say
   • updated alias current     env= function=say version=62
   • function updated          env= function=say name=lamb_say version=62
   • cleaning 1 functions     
   • hook                      command=rm -fr target env= function=say hook=clean
QA >> echo '{ "say": "Hello World!" }' | apex invoke say
   ⨯ Error: function response: Error loading class org.example.lambsay
QA >> echo '{ "say": "Hello World!" }' | apex invoke say -ldebug
   • loading functions in functions
   • open                      env= function=say
   ⨯ Error: function response: Error loading class org.example.lambsay

The work around is do the following:

cd functions/say/
lein uberjar
aws lambda create-function \ 
   --function-name clj-say \
   --handler org.example.lambsay \
   --runtime java8 \
   --memory 512 \
   --timeout 10 \
   --role arn:aws:iam::#########:role/test_lambda_function \
   --zip-file fileb://./target/org.example-0.0.0-SNAPSHOT-standalone.jar

Test cli-say function from AWS console, the function worked fine.  How sure how I can fix the apex deployment configuration.

Liath:
[...]

Issue #673, opened 2017-02-08
Stripping timestamps makes Clojure sad


Liath:
Because of how jars are repackaged currently, timestamp data is omitted. (The files are streamed out of the jar into the destination upload package  ([java.go:l93](https://github.com/apex/apex/blob/master/plugins/java/java.go#L93) & [zip.go:l42](https://github.com/apex/apex/blob/master/archive/zip.go#L42)))
Unfortunately, Clojure cares about timestamps and if `.clj` and their compiled `.class`s have different timestamps it will try to recompile them. This leads to extra load time and memory usage in Lambda. 

**To reproduce:**
 - Create a lein project in an apex function folder
 - Set function.json to `runtime: java` and `build-hook: "lein uberjar && mv target/-*-standalone.jar target/apex.jar"`
 - `apex build > test.zip`
 - Run this zip how Lambda will (close enough for our uses): `java -XX:MaxHeapSize=222823k -XX:MaxMetaspaceSize=26214k -XX:ReservedCodeCacheSize=13107k -XX:+UseSerialGC -XX:-TieredCompilation -jar test.zip`
 - Load hander in resulting clojure shell `(ns test (:require [com.test.what.ever :as test]))`
 - Presumably will fail with a message like `CompilerException java.lang.OutOfMemoryError: Metaspace, compiling:(a.clj:102:3)`

**There's a few ways to fix this:**
 - Pass a lein uberjar through instead of repackaging. ([example](https://github.com/Liath/apex/tree/clojure-support))
 - Add time stamps back. ([example](https://github.com/Liath/apex/tree/set-timestamps-addbytes))
 - Strip any clojure files we can find in java.go/Build ([example](https://github.com/Liath/apex/tree/strip-clojure-sources))

Idk what's best, but the last one seems most brittle yet easiest.

References:
http://stackoverflow.com/questions/10759343/make-lein-uberjar-without-source-files - Might work for small projects, but unfortunately many of my dependencies don't use it so probably useless.
http://stackoverflow.com/questions/19594360/preserving-timestamps-on-clojure-clj-files-when-building-shaded-jar-via-maven-s - More info on clojure's timestamp shenanigans.

tj:
Good to close this one now right?

Liath:
Yup! Closed. :)

PR #888, opened 2018-06-04
Allow for Plugins to specify a build artifact (fixes issue #887)

Liath:
We'd talked about doing this in the past (https://github.com/apex/apex/issues/673) and I was against this method then because it:
- bypasses `.apexignore` which could be confusing.
- bypasses the normal zip generation pipeline which makes interactions with other plugins more complicated if not impossible.
- breaks the hash check on deploy for clojure projects.

Unfortunately though, this is only method that will guarantee what get's deployed is what clojure users expect and I think that's more important.

Thoughts?

### DIFF
--- a/function/function.go
+++ b/function/function.go
@@ -99,6 +99,7 @@ type Config struct {
 	KMSKeyArn        string            `json:"kms_arn"`
 	DeadLetterARN    string            `json:"deadletter_arn"`
 	Zip              string            `json:"zip"`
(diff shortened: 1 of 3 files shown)


Liath:
The test message changes were a complaint from lint btw

Liath:
Hadn't seen the inference stuff before. I added `project.clj` for Clojure, which is a required file for compiling as far as I know.

Issue #883, opened 2018-05-08
Feature request: `region` field in project.json


astj:
I use multiple regions in a same AWS account to separate environments. (ex: ap-northeast-1 for production, us-east-1 for testing), and I'd like to handle those environments in apex's `--env` feature.

I can switch AWS _profiles_ using `profile` field in project.ENV.json, but `region` is not configurable in the JSON.

Currently I know some workarounds:
- Configure 2 AWS profiles for same account but different two regions, and specify the profile in project.ENV.json
  - like: `myaccount-tokyo` and `myaccount-virginia`
- Always set `apex --region` option carefully

But it's happy if I can set `region` field in project.ENV.json, like `profile`.

If this feature request is approved, I can open a p-r to implement this feature. (Ref: https://github.com/apex/apex/compare/master...astj:add-region-config?expand=1)

Liath:
Go ahead and create the PR. It seems reasonable enough to me and you're not the first to ask for this.

astj:
@Liath Okay. I opened #889.

astj:
I've just found #824.  Though project-wide region configuration is better in my project, but that can be an alternative, by specifying region in function.ENV.json of each functions.

tj:
merged, thanks guys!

PR #889, opened 2018-06-05
add `region` field to project configuration

astj:
From #883 

Add `region` field to project.json.

The priority of region specification will be following:
(Most Prior) `--region` > project.json 🆕 > `AWS_REGION` env > `~/.aws/config` (Least prior)


tj:
thanks!

PR #925, opened 2018-12-26
install.sh: retry failed downloads

### LINKED ISSUE #924: Installation fails occasionally with 'Empty reply from server', add retries?
If you have a bug report please let us know the following:

  - __Operating system__: Linux/Ubuntu on CircleCI
  - __Apex Version__: Not relevant

I'm installing Apex with:

```
    curl https://raw.githubusercontent.com/apex/apex/master/install.sh | sh
```

Occasionally, I get:

```
apex/apex: checking GitHub for latest version
apex/apex: found version 1.0.0-rc3 for linux/amd64
apex/apex: downloading https://github.com/apex/apex/releases/download/v1.0.0-rc3/apex_1.0.0-rc3_linux_amd64.tar.gz
curl: (52) Empty reply from server
```

Maybe this is an issue with too many connections from CircleCI to the Github IP address pool. Still, it seems helpful to try to retry the download in the event of failure.

kevinburke:
Occasionally, downloading the Apex binary fails with a "Empty reply"
error from Github. Since the install script is only issuing GET
requests these should be safe to retry, and we can increase download
reliability by doing so.

The `--retry` flag was added in version 7.12.3, released in December
2004, so it should be available on most systems that are downloading
Apex.

Fixes #924.

Before sending a pull-request please open an issue to discuss new features or non-trivial changes.

Thanks for your contribution!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -194,7 +194,7 @@ http_download() {
   headerflag=''
   destflag=''
   if is_command curl; then
-    cmd='curl --fail -sSL'
+    cmd='curl --retry 3 --fail -sSL'
     destflag='-o'
     headerflag='-H'
   elif is_command wget; then


tj:
hmm weird that it's failing like that from Github, sounds good to me though thanks!

